PR #21, opened 2026-03-18
feat(T4Dataset): add initial T4dataset generator 

KSeangTan:
## Summary
This PR adds an initial dataset interface/protocol for dataset generator, where it will preprocess datasets and make them a format consumed by downstream training/evaluation tasks. This PR only define interface and implements the basic table format for `T4Dataset`, and more columns/metadata will be added in the next PR.  The generation starts from `autoware_ml/scripts/generate_dataset.py`. For more details, please check the README here:

- [Database](https://github.com/KSeangTan/autoware-ml/blob/54daf61d612138800e8688af00bb42dd0871e3df/autoware_ml/databases/README.md)
- [T4Database](https://github.com/KSeangTan/autoware-ml/blob/54daf61d612138800e8688af00bb42dd0871e3df/autoware_ml/databases/t4datasets/README.md)

Specifically, it makes the following changes:
- Create `uv.lock` to sync envinroment installation [[1]](https://github.com/KSeangTan/autoware-ml/blob/81f5c74ac7be77a3a8bf801de609fc1c774259a4/uv.lock)
- Update `setup.py` to import `autoware_ml` before actual installation [[2]](https://github.com/tier4/autoware-ml/pull/21/changes#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7R3-R8)
- Add `polars` and `pydantic` to the installation [[3]](https://github.com/tier4/autoware-ml/pull/21/changes#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R39-R40)
- Add more ignore words [[4]](https://github.com/tier4/autoware-ml/pull/21/changes#diff-4e0f175c840784de9e935f881821905c75a2a125aa57c51268a7fb4496edfbe6R148-R153)
- Add `autoware_ml/scripts/generate_dataset.py` to preprocess datasets. Noe that this script will be removed in another PR once the dataloader is ready. [[5]](https://github.com/tier4/autoware-ml/pull/21/changes#diff-25d4faa0e9c92a781f90bb808c6f5cf852f5559cf55ba5cd5f034a0d31b20df8R1-R64) 
[...]

PR #33, opened 2026-04-07
chore(pixi): add polars and pydantic to pixi

KSeangTan:
## Summary

This PR updates pixi by adding `polars` and `pydantic` to the dependency, and it fixes the local installation error if `autoware-ml` is not installed as a library by appending `autoware-ml` to the system path.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,6 +40,8 @@ dependencies = [
     "torch==2.9.1+cu128",
     "torchvision==0.24.1+cu128",
     "transforms3d==0.4.2",
+    "polars>=1.38.1,<2.0.0",
+    "pydantic>=2.12.5,<2.13.0"
 ]
 
 [project.optional-dependencies]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,6 +31,8 @@ dependencies = [
     "onnxscript==0.6.2",
     "opencv-python-headless==4.11.0.86",
     "optuna-dashboard==0.20.0",
+    "polars>=1.38.1,<2.0.0",
+    "pydantic>=2.12.5,<2.13.0",
     "pytest==9.0.2",
     "rerun-sdk==0.23.1",
     "scipy==1.17.1",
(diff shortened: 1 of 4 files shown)


setup.py
mojomex, line 1: We shouldn't have a `setup.py` anymore. `pyproject.toml` should define project packages, then the project will just work without `setup.py`.

mojomex: (can be dealt with in a separate PR though!)

mojomex (approved):

setup.py
mojomex: Nevermind, see https://github.com/tier4/autoware-ml/pull/21#discussion_r3043784146

pyproject.toml
```diff
@@ -40,6 +40,8 @@ dependencies = [
     "torch==2.9.1+cu128",
     "torchvision==0.24.1+cu128",
     "transforms3d==0.4.2",
+    "polars>=1.38.1,<2.0.0",
```
amadeuszsz: Please keep the alphabetical order.

setup.py
```diff
@@ -1,7 +1,11 @@
 """Setup entrypoint for Autoware-ML with compiled extensions."""
 
+import sys
+import os
 from setuptools import setup
 
+# Add the project root directory to the Python path, so that the build script can be found
+sys.path.append(os.path.dirname(os.path.abspath(__file__)))
 from autoware_ml.ops.build import get_cmdclass, get_ext_modules
```
amadeuszsz: Maybe it's only me, but if possible, I'm trying to avoid any code between imports. It also breaks linters behavior.

Could you please check this impl?
```python
"""Setup entrypoint for Autoware-ML with compiled extensions."""

from setuptools import setup
from importlib.util import module_from_spec, spec_from_file_location
from pathlib import Path

build_path = Path(__file__).parent / "autoware_ml" / "ops" / "build.py"
spec = spec_from_file_location("autoware_ml_setup_build", build_path)
module = module_from_spec(spec)
assert spec is not None and spec.loader is not None
spec.loader.exec_module(module)

setup(
    ext_modules=module.get_ext_modules(),
    cmdclass=module.get_cmdclass(),
)
```

KSeangTan: It works and changed here:
59cc3723141962574f84c561ed7e290eebed512e

pyproject.toml
KSeangTan: make sense, updated here:
cf505d24e53f9a6f5dc0159825efa76d3a195c67

amadeuszsz (approved):
LGTM

PR #47, opened 2026-06-10
refactor(deploy): allow export specs to provide dynamic axes

mojomex:
## Summary
- Add optional `ExportSpec.dynamic_axes` for legacy ONNX exports.
- Make deploy export prefer spec-generated dynamic axes over YAML dynamic axes when `dynamo=false`.
- Add a unit test covering the override behavior.

## Why
Some export modules generate their input contract in code. Letting the same export spec provide dynamic axes keeps generated input names and generated dynamic axes together, avoiding brittle duplicated YAML.

For example, PTv3 in PR #46 generates onnx inputs and dynamic axes dependent on the number of encoder stages, so a static YAML is brittle.

## Verification
- `pixi run --environment dev pytest autoware_ml/tests/utils/test_deploy.py -q`
- `pixi run --environment dev pre-commit run --files autoware_ml/utils/deploy.py autoware_ml/scripts/deploy.py autoware_ml/tests/utils/test_deploy.py`

### DIFF
--- a/autoware_ml/scripts/deploy.py
+++ b/autoware_ml/scripts/deploy.py
@@ -222,6 +222,7 @@ def main(cfg: DictConfig) -> None:
                         module_onnx_cfg,
                         export_spec.input_param_names,
                         export_spec.output_names,
+                        export_spec.dynamic_axes,
                         module_onnx_path,
                     )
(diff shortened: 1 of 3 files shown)


amadeuszsz (approved):
LGTM

PR #73, opened 2026-07-16
feat(models): add StreamPETR

amadeuszsz:
## Summary

- Add StreamPETR, a camera-only temporal 3D detector: VoVNet-99 multi-scale backbone, geometry-aware query initialization, propagated temporal memory bank with hybrid self-attention, and denoising query training
- Provide t4dataset (480x640) and nuScenes (320x800) configs with scene-streaming dataloading: `GroupStreamingSampler` feeds scene-contiguous frames per batch lane, evaluation runs one lane per rank so every frame is scored
- Train with `bf16-mixed` precision using float32 islands for frustum geometry, attention, prediction branches, and temporal-memory ego-pose math; frame timestamps stay float64 end to end
- Export three ONNX modules (`extract_img_feat`, `position_embedding`, `pts_head_memory`) matching the `autoware_camera_streampetr` runtime interface, traced on CPU for device-neutral graphs
- Add a shared multiview detection datamodule base with the streaming dataloader policy and dtype-preserving scalar collation
- Add the `normalize_boxes3d`/`denormalize_boxes3d` box-encoding pair in `bbox_coders` shared by assignment, loss, and decoding, plus a full-box L1 Hungarian matching cost (`BBox3DL1Cost`)
- Rework `ResizeCropFlipRotImage` so the affine transform matches the pixel mapping exactly (single resize, zero-padded crop, random horizontal crop in training); evaluation behavior of existing consumers is unchanged
- Add in-model `GridMask` augmentation and grouped optimizer overrides with a reduced backbone learning rate
- Validate required stream metadata at the model boundary and fail loudly when the datamodule does not stream scene-contiguous batches
- Cover forward/loss/export ABI, DDP loss-key uniformity, bf16 memory-motion, float64 timestamp propagation, and resize-affine consistency with unit tests

## Pretrained initialization

Base checkpoint: nuScenes-trained StreamPETR VoV-99 from the TIER IV model zoo:
https://download.autoware-ml-model-zoo.tier4.jp/autoware-ml/models/streampetr/streampetr-vov99/nuscenes/v1.0/nuscenes_vov99_baseline_320x800.pth

The checkpoint was converted offline (one-off script, not part of this PR) to [streampetr_vov99_nuscenes_pretrain_converted.pth](https://drive.google.com/drive/folders/11rtIg1fDKPNBG9x6jcwXQ1LL9CB_9VDM?usp=sharing) and is loaded with `train --weights`. Conversion steps:

- Renamed modules to this repo's naming: `pts_bbox_head` -> `bbox_head`; decoder layers `attentions.0.attn` -> `self_attn`, `attentions.1.attn` -> `cross_attn`, `ffns.0.layers.0.0`/`.1` -> `ffn.0`/`.3`, `norms.0/1/2` -> `norm1/2/3`, `transformer.decoder.post_norm` -> `post_norm`; `featurized_pe.conv_reduce`/`conv_expand` -> `reduce`/`expand`
- Backbone keys transfer 1:1 (626/626); the stem conv's input channels are flipped because the source model consumes BGR images while this repo's pipeline is RGB (image normalization stats are RGB-ordered in the configs accordingly)
- Neck weights are derived analytically: the two `CPFPN` lateral 1x1 convs concatenate along input channels into the fused `GeneralizedLSSFPN` lateral conv (pointwise convs commute with upsampling), conv biases are absorbed into identity-initialized BatchNorm, and the 3x3 fpn conv maps to the output conv
- Regression head layout is identical (`[cx, cy, cz, log dims, sin, cos, vx, vy]`); no channel permutation needed
- Dropped tensors: the 10-class classification output layer (this model uses the 7-class t4 taxonomy; it trains from scratch), head constant buffers (`code_weights`, `match_costs`, `pc_range`, `position_range`, `coords_d`), and the 2D auxiliary `img_roi_head` absent from this port
- Backbone/neck keys are duplicated into the `image_feature_extractor` namespace so coverage reports stay clean

Every model tensor except the final classification layer is initialized from the source checkpoint.

## Checklist


KSeangTan:
Let's keep @tzhong518 in the loop

PR #78, opened 2026-07-24
fix(datamodule): store annotations in a fork-shared serialized buffer to stop DataLoader OOM

vividf:
## Problem

During multi-GPU StreamPETR training (2 GPUs × `num_workers=32`), training was unexpectedly terminated by the Linux OOM killer:

```text
[rank: 1] Child process ... terminated with code -9
```

This was **not a CUDA OOM**, but a **system RAM exhaustion** issue.

**Note**: The persistence worker is already set to **False** when this issue happened. A quick fix would be decreasing the number of workers.

### Root Cause

Each dataset stored annotations as a standard Python:

```python
self.data_infos: list[dict]
```

When `DataLoader(num_workers > 0)` is used on Linux, PyTorch launches worker
processes via **fork()**.

Initially, parent and child processes share memory through **copy-on-write (CoW)**,
so the dataset should theoretically occupy memory only once.

However, CPython stores the **reference count (`ob_refcnt`) inside every Python
object**. Every time a worker merely **reads** an element from
`list[dict]`, CPython increments and decrements its reference count,
which modifies the object's memory page.

As a result, the shared page becomes dirty and CoW creates a private copy for
that worker.

Therefore, each DataLoader worker gradually duplicates the annotation dataset,
even though it only performs read operations.

### Why It Became Fatal

The j6gen2 training annotation file is:

- **941 MB** on disk
- Approximately **5 GB** after being loaded into Python objects

With:

- 2 GPUs
- `num_workers = 32`

there are **64 worker processes**.

Worst-case memory usage becomes approximately:

```
64 × 5 GB ≈ 320 GB
```

The shared training machine only had **125 GB RAM**, so during epoch 3 the
kernel OOM killer terminated the training process.

## Solution

A new container class was introduced:

```
autoware_ml/datamodule/common/serialization.py
```

## `SerializedSampleList`

The implementation mirrors mmengine's serialization strategy.

Instead of storing:

```text
list
 ├── dict
 ├── dict
 ├── dict
 └── ...
```

samples are converted into:

```
+------------------------------+
| Serialized byte buffer       |
+------------------------------+

+------------------------------+
| int64 offset table           |
+------------------------------+
```

### Access Flow

```text
Dataset
      │
      ▼
SerializedSampleList
      │
      ├── contiguous pickle buffer
      ├── offset table
      ▼
__getitem__(i)
      │
      ▼
pickle.loads(buffer[start:end])
      ▼
fresh Python dict
```

Each sample is deserialized **only when accessed**.

Since the underlying storage is a byte array rather than millions of Python
objects, worker processes never modify shared pages during iteration.

The storage therefore remains shared across all forked workers.

### Properties

- Drop-in replacement for `list`
- Supports:
  - `len()`
  - integer indexing
- Returns independent Python objects
- No shared mutable state between workers

---

## Dataset Migration

The new container was applied to every dataset previously storing
`list[dict]`.

Serialization is performed **as the final step of `__init__`**, after all
operations requiring a full dataset traversal are complete.

These include:

- filtering
- `frame_weights`
- `prev_exists`
- any other preprocessing passes

## Updated datasets

- `common/multiview_detection3d.py`
  - additionally caches `scene_index_groups()` during initialization because
    samplers access it after worker processes are created
- `t4dataset/detection3d.py`
- `nuscenes/detection3d.py`
- `t4dataset/segmentation3d.py`
- `nuscenes/segmentation3d.py`
- `t4dataset/segdet.py`
- `nuscenes/segdet.py`
- `t4dataset/calibration_status.py`

---

### Unit Tests

New test file:

```
tests/datamodule/test_serialization.py
```

Coverage includes:

- serialization/deserialization round-trip
- equality with original data
- negative indexing
- out-of-bound indexing
- iteration
- copy independence
- pickle round-trip
- forked worker shared-read behavior

---

### Test Results

```
542 tests passed
```

---

## Copy-on-Write Memory Measurement

The real validation dataset was used for measurement.

Dataset statistics:

- 3,215 frames
- approximately 270 MB live Python objects

Experiment:

- 4 forked worker processes
- every worker reads every sample twice
- garbage collection forced
- `Private_Dirty` measured using:

```
/proc/self/smaps_rollup
```

### Results

| Storage Layout | Private Memory per Worker |
|----------------|--------------------------:|
| `list[dict]` | 230 MB |
| `SerializedSampleList` | 7.4 MB |

The original implementation privatized roughly **85% of the dataset** for every
worker.

The serialized implementation remains nearly constant regardless of dataset size.

---

## Memory Reduction

| Metric | Result |
|---------|--------|
| Private memory per worker | **230 MB → 7.4 MB** |
| Reduction | **31× smaller** |

Extrapolating to the 5 GB training annotations:

| Layout | Estimated Memory @ 64 Workers |
|--------|------------------------------:|
| Original `list[dict]` | ~275–320 GB |
| `SerializedSampleList` | Nearly constant |

The estimate closely matches the observed Linux OOM event, while the serialized
implementation eliminates the copy-on-write memory explosion and allows
high-worker-count training to run successfully.

### DIFF
--- a/autoware_ml/datamodule/common/multiview_detection3d.py
+++ b/autoware_ml/datamodule/common/multiview_detection3d.py
@@ -33,6 +33,7 @@ from autoware_ml.datamodule.common.detection3d import (
     build_label_to_category,
     load_detection_data_infos,
 )
+from autoware_ml.datamodule.common.serialization import SerializedSampleList
 
 logger = logging.getLogger(__name__)
 
@@ -110,11 +111,14 @@ class MultiviewDetection3DDataset(Dataset):
         self.require_image_files = require_image_files
         with open(ann_file, "rb") as file:
             data = pickle.load(file)
-        self.data_infos = load_detection_data_infos(data)
+        data_infos = load_detection_data_infos(data)
         if filter_frames_with_camera_order:
-            self.data_infos = self._filter_frames_with_camera_order(self.data_infos)
-        self.prev_exists = self._build_prev_exists(self.data_infos)
+            data_infos = self._filter_frames_with_camera_order(data_infos)
+        self.prev_exists = self._build_prev_exists(data_infos)
         self.label_to_category = build_label_to_category(data.get("metainfo", {}))
+        # Serialize last: the full passes above must run on the live list.
+        self._scene_index_groups = self._build_scene_index_groups(data_infos)
+        self.data_infos = SerializedSampleList(data_infos)
 
     @staticmethod
     def _build_prev_exists(data_infos: list[dict[str, Any]]) -> np.ndarray:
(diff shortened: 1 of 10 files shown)


amadeuszsz:
@codex review

amadeuszsz (approved):
LGTM!

PR #82, opened 2026-08-04
fix(datamodule): fix datamodule for sweep transformation

vividf:
## Description
`build_sweep_entries` in `autoware_ml/datamodule/common/detection3d.py`
recomposed the sweep → key-lidar transform from stored ego poses:
sweep2key = inv(key_ego2global @ key_lidar2ego) @ sweep_ego2global @ sweep_lidar2ego

This is correct only when the info file's sweep `lidar2ego` is the static
mounting extrinsic. Existing AWML-generated info files incorrectly stored the
sweep → key transform (which already contains ego motion) in that field, so
the composition applied ego motion twice and sweep point clouds were
misaligned — worse at higher ego speed.

## Fix

Prefer the precomputed `lidar2sensor` matrix (key → sweep, ego motion
included) stored per sweep and invert it to get `sweep2key`. This field has
always been correct, so the datamodule now produces aligned sweeps with both
old (buggy `lidar2ego`) and regenerated info files. Recomposing from ego
poses is kept only as a fallback when `lidar2sensor` is absent, and the
key-frame inverse is computed lazily in that path.

Related: AWML-side fix that corrects the `lidar2ego` field at info-generation
time https://github.com/tier4/AWML/pull/223). Either change alone fixes the misalignment; this one
additionally keeps existing pickle files usable without regeneration.

## Checklist

### DIFF
--- a/autoware_ml/datamodule/common/detection3d.py
+++ b/autoware_ml/datamodule/common/detection3d.py
@@ -76,8 +76,11 @@ def build_sweep_entries(sample: Mapping[str, Any]) -> list[dict[str, Any]]:
     """Convert stored ``lidar_sweeps`` metadata into loader-ready sweep entries.
 
     Each entry carries the sweep point-cloud path, its timestamp, and the
-    rigid transform from the sweep lidar frame into the key lidar frame,
-    composed from the stored ego poses and lidar extrinsics.
+    rigid transform from the sweep lidar frame into the key lidar frame.
+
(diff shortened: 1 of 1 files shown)


amadeuszsz:
@codex review

amadeuszsz (approved):
LGTM, thank you!

As a next time we should clarify dataset contract, hopefully shared across all the datasets. I would see it as a key(required/optional)-description table in our documentation.

PR #88, opened 2026-08-06
feat(bevfusion): correct behaviors that deviated from the reference implementation

vividf:
## What

Fixes a set of silent deviations from the reference (mmdet3d) BEVFusion that
were introduced during the migration. Found by running the same reference
BEVFusion 2.8 checkpoint through both frameworks and chasing every output
difference to its root cause.

| Fix | Before | After | Evidence |
|---|---|---|---|
| Decoder positional embeddings | baked into the tokens once; the residual stream carries them | DETR/TransFusion semantics: residual stream stays position-free, embeddings added to q/k/v at every attention call | Not an equivalent function — with identical weights the old math produced 354 boxes where the reference produces 113 |
| `remove_close` region | radial circle | axis-aligned box `\|x\|,\|y\| < radius` | nuScenes-devkit `PointCloud.remove_close` is a box (its docstring says "radius", the code is a box); mmdet3d/OpenPCDet/CenterPoint all inherit it — the circle came from implementing the docstring |
| Circle-NMS radius (j6gen2) | 0.25 | 0.5 (+ comment) | mmdet3d's `circle_nms` threshold is a **squared** distance, so its 0.25 means a 0.5 m Euclidean radius — the value was copied without translating the semantics, halving the intended suppression radius |
| `dense_heatmap_pooling_classes` (j6gen2) | car, truck, bus | + barrier | the trained reference recipe includes barrier |
| Prediction-head hidden width | hardcoded `= in_channels` (128) | configurable (`head_hidden_channels`), j6gen2 sets the reference value 64 | mmdet3d's `SeparateHead` has this degree of freedom (`head_conv`, default 64); the port lost the parameter and silently doubled the width |
| Head BN epsilon | hardcoded 1e-3 | configurable (`norm_eps`), j6gen2 sets 1e-5 | 1e-3 is an mmcv-ism; the reference heads use the PyTorch default 1e-5 |
| Sweep selection at train time | always nearest | `selection: random` in the bevfusion train pipelines | matches mmdet3d's train-time behavior: temporal-jitter augmentation (robustness to frame drops) that also keeps the per-point time-lag channel informative |
| Voxel budget at evaluation | single `max_voxels` 120k for train and eval | `eval_max_voxels` 160k (train stays 120k) | typical j6gen2 val frames occupy 120k–136k voxels, so evaluation silently truncated ~10% of them; the reference uses `max_voxels=[120000, 160000]` |

Supporting change: the (deliberately unregistered) `DataPreprocessing`
pipeline now mirrors the owning model's train/eval mode — without this,
`model.eval()` never reached the preprocessor and any train/eval-dependent
preprocessing was silently stuck in training mode.

All new parameters default to the previous behavior; only the bevfusion
configs opt into the corrected values. Existing checkpoints of other tasks are
unaffected (note: any checkpoint trained with the old decoder math would
change behavior on the new code — none is known to be in use).

## Verification

- Root causes were isolated with a weight-level checkpoint conversion study
  (AWML BEVFusion 2.8 → autoware-ml): with these fixes (plus temporary shims
  for the two remaining intentional architecture differences — shared-conv
  norm/act and heatmap border handling), the reference checkpoint reproduces
  113/113 boxes on the parity sample and val mAP/mAPH matching the reference
  evaluation to ±0.0004.
- Occupied-voxel counts measured on j6gen2 val confirm the evaluation
  truncation (5 samples: 113k / 120k / 131k / 136k / 132k vs the 120k budget).
- Training smoke test on the j6gen2 lidar config (10 train + 2 val batches,
  batch size 4, single GPU): runs end to end with finite losses
  (train/loss 18.6 at start), exercising the random sweep selection, the
  64-wide heads, the corrected decoder math, and the eval voxel budget.

The end-to-end verification numbers above were produced together with #82
(sweep transform fix) and #84 (gt_num_points collation fix); this PR is
independent of both at the code level.

### DIFF
--- a/autoware_ml/configs/tasks/detection3d/bevfusion/camera_lidar_voxel0075_second_secfpn_54m_nuscenes.yaml
+++ b/autoware_ml/configs/tasks/detection3d/bevfusion/camera_lidar_voxel0075_second_secfpn_54m_nuscenes.yaml
@@ -55,6 +55,7 @@ datamodule:
         time_dim: 4
         pad_empty_sweeps: true
         remove_close: true
+        selection: random
       - _target_: autoware_ml.transforms.camera_lidar.camera_lidar.ImageAug3D
         final_dim: ${image_size}
         resize_lim: [0.38, 0.55]
--- a/autoware_ml/configs/tasks/detection3d/bevfusion/camera_lidar_voxel0075_second_secfpn_54m_nuscenes.yaml
+++ b/autoware_ml/configs/tasks/detection3d/bevfusion/camera_lidar_voxel0075_second_secfpn_54m_nuscenes.yaml
@@ -55,7 +55,7 @@ datamodule:
         time_dim: 4
         pad_empty_sweeps: true
         remove_close: true
-        selection: random
+        test_mode: false
       - _target_: autoware_ml.transforms.camera_lidar.camera_lidar.ImageAug3D
         final_dim: ${image_size}
         resize_lim: [0.38, 0.55]
--- a/autoware_ml/configs/tasks/detection3d/bevfusion/camera_lidar_voxel0170_second_secfpn_120m_t4dataset_j6gen2.yaml
+++ b/autoware_ml/configs/tasks/detection3d/bevfusion/camera_lidar_voxel0170_second_secfpn_120m_t4dataset_j6gen2.yaml
@@ -61,6 +61,7 @@ datamodule:
         time_dim: 4
         pad_empty_sweeps: true
         remove_close: true
+        selection: random
       - _target_: autoware_ml.transforms.camera_lidar.camera_lidar.ImageAug3D
         final_dim: ${image_size}
         resize_lim: [0.29, 0.35]
--- a/autoware_ml/configs/tasks/detection3d/bevfusion/camera_lidar_voxel0170_second_secfpn_120m_t4dataset_j6gen2.yaml
+++ b/autoware_ml/configs/tasks/detection3d/bevfusion/camera_lidar_voxel0170_second_secfpn_120m_t4dataset_j6gen2.yaml
@@ -61,7 +61,7 @@ datamodule:
         time_dim: 4
         pad_empty_sweeps: true
         remove_close: true
-        selection: random
+        test_mode: false
       - _target_: autoware_ml.transforms.camera_lidar.camera_lidar.ImageAug3D
         final_dim: ${image_size}
         resize_lim: [0.29, 0.35]
--- a/autoware_ml/datamodule/common/detection3d.py
+++ b/autoware_ml/datamodule/common/detection3d.py
@@ -76,8 +76,11 @@ def build_sweep_entries(sample: Mapping[str, Any]) -> list[dict[str, Any]]:
     """Convert stored ``lidar_sweeps`` metadata into loader-ready sweep entries.
 
     Each entry carries the sweep point-cloud path, its timestamp, and the
-    rigid transform from the sweep lidar frame into the key lidar frame,
-    composed from the stored ego poses and lidar extrinsics.
+    rigid transform from the sweep lidar frame into the key lidar frame.
+
+    The precomputed ``lidar2sensor`` matrix (the key-lidar → sweep-lidar
+    transform including ego motion) is preferred when present. Recomposing the
+    transform from the stored ego poses is only a fallback
 
     Args:
         sample: Raw sample dictionary with ``lidar_points``, ``ego2global``,
@@ -94,15 +97,21 @@ def build_sweep_entries(sample: Mapping[str, Any]) -> list[dict[str, Any]]:
     if not lidar_sweeps:
         return []
 
-    key_lidar2ego = np.asarray(sample["lidar_points"]["lidar2ego"], dtype=np.float64)
-    key_ego2global = np.asarray(sample["ego2global"], dtype=np.float64)
-    global2key_lidar = np.linalg.inv(key_ego2global @ key_lidar2ego)
+    global2key_lidar = None
 
     entries = []
     for sweep in lidar_sweeps:
-        sweep_lidar2ego = np.asarray(sweep["lidar_points"]["lidar2ego"], dtype=np.float64)
-        sweep_ego2global = np.asarray(sweep["ego2global"], dtype=np.float64)
-        sweep2key_lidar = global2key_lidar @ sweep_ego2global @ sweep_lidar2ego
+        lidar2sensor = sweep["lidar_points"].get("lidar2sensor")
+        if lidar2sensor is not None:
+            sweep2key_lidar = np.linalg.inv(np.asarray(lidar2sensor, dtype=np.float64))
+        else:
+            if global2key_lidar is None:
+                key_lidar2ego = np.asarray(sample["lidar_points"]["lidar2ego"], dtype=np.float64)
+                key_ego2global = np.asarray(sample["ego2global"], dtype=np.float64)
+                global2key_lidar = np.linalg.inv(key_ego2global @ key_lidar2ego)
+            sweep_lidar2ego = np.asarray(sweep["lidar_points"]["lidar2ego"], dtype=np.float64)
+            sweep_ego2global = np.asarray(sweep["ego2global"], dtype=np.float64)
+            sweep2key_lidar = global2key_lidar @ sweep_ego2global @ sweep_lidar2ego
         entries.append(
             {
                 "lidar_path": sweep["lidar_points"]["lidar_path"],
--- a/autoware_ml/models/base.py
+++ b/autoware_ml/models/base.py
@@ -104,6 +104,9 @@ class BaseModel(MetricEvalMixin, L.LightningModule, ABC):
             Batch dictionary after runtime preprocessing.
         """
         del dataloader_idx
+        # The preprocessing pipeline is not a registered submodule, so keep its
+        # train/eval mode in sync with the model before applying it.
+        self._data_preprocessing.train(self.training)
         return self._data_preprocessing(batch_inputs_dict)
 
     def predict_outputs(self, batch_inputs_dict: Mapping[str, Any], outputs: Any) -> Any:
--- a/autoware_ml/models/base.py
+++ b/autoware_ml/models/base.py
@@ -104,10 +104,7 @@ class BaseModel(MetricEvalMixin, L.LightningModule, ABC):
             Batch dictionary after runtime preprocessing.
         """
         del dataloader_idx
-        # The preprocessing pipeline is not a registered submodule, so keep its
-        # train/eval mode in sync with the model before applying it.
-        self._data_preprocessing.train(self.training)
-        return self._data_preprocessing(batch_inputs_dict)
+        return self._data_preprocessing(batch_inputs_dict, is_training=self.training)
 
     def predict_outputs(self, batch_inputs_dict: Mapping[str, Any], outputs: Any) -> Any:
         """Convert raw model outputs into task-level predictions.
--- a/autoware_ml/models/common/layers/conv.py
+++ b/autoware_ml/models/common/layers/conv.py
@@ -29,6 +29,7 @@ class ConvModule(nn.Module):
         out_channels: int,
         stride: int = 1,
         transpose: bool = False,
+        norm_eps: float = 1e-3,
     ) -> None:
         """Initialize the convolution block.
 
@@ -37,6 +38,7 @@ class ConvModule(nn.Module):
             out_channels: Output channel count.
             stride: Convolution stride.
             transpose: Whether to use transposed convolution.
+            norm_eps: Epsilon used by the batch normalization layer.
         """
         super().__init__()
         if transpose:
@@ -56,7 +58,7 @@ class ConvModule(nn.Module):
                 padding=1,
                 bias=False,
             )
-        self.norm = nn.BatchNorm2d(out_channels, eps=1e-3, momentum=0.01)
+        self.norm = nn.BatchNorm2d(out_channels, eps=norm_eps, momentum=0.01)
         self.activation = nn.ReLU(inplace=True)
 
     def forward(self, x: torch.Tensor) -> torch.Tensor:
--- a/autoware_ml/models/common/layers/conv.py
+++ b/autoware_ml/models/common/layers/conv.py
@@ -30,6 +30,7 @@ class ConvModule(nn.Module):
         stride: int = 1,
         transpose: bool = False,
         norm_eps: float = 1e-3,
+        norm_momentum: float = 0.01,
     ) -> None:
         """Initialize the convolution block.
 
@@ -39,6 +40,7 @@ class ConvModule(nn.Module):
             stride: Convolution stride.
             transpose: Whether to use transposed convolution.
             norm_eps: Epsilon used by the batch normalization layer.
+            norm_momentum: Momentum used by the batch normalization layer.
         """
         super().__init__()
         if transpose:
@@ -58,7 +60,7 @@ class ConvModule(nn.Module):
                 padding=1,
                 bias=False,
             )
-        self.norm = nn.BatchNorm2d(out_channels, eps=norm_eps, momentum=0.01)
+        self.norm = nn.BatchNorm2d(out_channels, eps=norm_eps, momentum=norm_momentum)
         self.activation = nn.ReLU(inplace=True)
 
     def forward(self, x: torch.Tensor) -> torch.Tensor:
--- a/autoware_ml/models/detection3d/heads/transfusion.py
+++ b/autoware_ml/models/detection3d/heads/transfusion.py
@@ -58,21 +58,22 @@ class LearnedPositionalEncoding(nn.Module):
         """
         super().__init__()
         self.proj = nn.Sequential(
-            nn.Linear(input_channels, embed_dims),
+            nn.Conv1d(input_channels, embed_dims, kernel_size=1),
+            nn.BatchNorm1d(embed_dims),
             nn.ReLU(inplace=True),
-            nn.Linear(embed_dims, embed_dims),
+            nn.Conv1d(embed_dims, embed_dims, kernel_size=1),
         )
 
     def forward(self, position: torch.Tensor) -> torch.Tensor:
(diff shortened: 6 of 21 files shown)


amadeuszsz:
@codex review

autoware_ml/configs/tasks/detection3d/bevfusion/camera_lidar_voxel0075_second_secfpn_54m_nuscenes.yaml
```diff
@@ -55,6 +55,7 @@ datamodule:
         time_dim: 4
         pad_empty_sweeps: true
         remove_close: true
+        selection: random
```
KSeangTan: Use `test_mode` instead of `selection`

autoware_ml/configs/tasks/detection3d/bevfusion/camera_lidar_voxel0170_second_secfpn_120m_t4dataset_j6gen2.yaml
```diff
@@ -61,6 +61,7 @@ datamodule:
         time_dim: 4
         pad_empty_sweeps: true
         remove_close: true
+        selection: random
```
KSeangTan: same as above

autoware_ml/datamodule/common/detection3d.py
```diff
@@ -94,15 +97,21 @@ def build_sweep_entries(sample: Mapping[str, Any]) -> list[dict[str, Any]]:
...
-    key_ego2global = np.asarray(sample["ego2global"], dtype=np.float64)
-    global2key_lidar = np.linalg.inv(key_ego2global @ key_lidar2ego)
+    global2key_lidar = None
 
     entries = []
     for sweep in lidar_sweeps:
-        sweep_lidar2ego = np.asarray(sweep["lidar_points"]["lidar2ego"], dtype=np.float64)
-        sweep_ego2global = np.asarray(sweep["ego2global"], dtype=np.float64)
-        sweep2key_lidar = global2key_lidar @ sweep_ego2global @ sweep_lidar2ego
+        lidar2sensor = sweep["lidar_points"].get("lidar2sensor")
```
KSeangTan, line 104: Move this part into another function, and make the separation cleaner.

autoware_ml/models/common/layers/conv.py
```diff
@@ -56,7 +58,7 @@ def __init__(
                 padding=1,
                 bias=False,
             )
-        self.norm = nn.BatchNorm2d(out_channels, eps=1e-3, momentum=0.01)
+        self.norm = nn.BatchNorm2d(out_channels, eps=norm_eps, momentum=0.01)
```
KSeangTan: If we want to make this configurable, might as well add the changes to `momentum`.

autoware_ml/models/detection3d/heads/transfusion.py
```diff
@@ -127,15 +127,26 @@ def forward(
...
+        query_pos_embed = self.query_pos_encoding(query_pos)
+        key_pos_embed = self.key_pos_encoding(key_pos)
 
-        attended, _ = self.self_attn(query_tokens, query_tokens, query_tokens)
+        attended, _ = self.self_attn(
+            query_tokens + query_pos_embed,
+            query_tokens + query_pos_embed,
+            query_tokens + query_pos_embed,
+        )
         query_tokens = self.norm1(query_tokens + self.dropout(attended))
```
KSeangTan, line 144: It seems that the implementation in mmdet doesn't apply dropout to attended values:
https://github.com/open-mmlab/mmdetection/blob/main/mmdet/models/layers/transformer/detr_layers.py#L352-L372

https://github.com/tier4/AWML/blob/main/projects/BEVFusion/bevfusion/transformer.py#L86-L110

However, our implementation is more closely to Pytorch's implementation. Suggest to have three sets of dropout values: 
```
self.mha_dropout = nn.Dropout(mha_dropout)
self.dropout_after_attention = nn.Dropout(dropout_after_attention)
self.ffn_dropout = nn.Dropout(ffn_dropout)
```

autoware_ml/models/detection3d/heads/transfusion.py
```diff
@@ -127,15 +127,26 @@ def forward(
...
             Refined query tensor.
         """
+        # TransformerDecoderLayer:
+        # The residual stream stays position-free
+        # and the positional embeddings are added to q/k/v at every attention call.
         query_tokens = query.transpose(1, 2)
         key_tokens = key.transpose(1, 2)
-        query_tokens = query_tokens + self.query_pos_encoding(query_pos)
-        key_tokens = key_tokens + self.key_pos_encoding(key_pos)
+        query_pos_embed = self.query_pos_encoding(query_pos)
```
KSeangTan, line 136: Both pos encoding embeddings have batchnorm after the linear layer:
https://github.com/tier4/AWML/blob/main/projects/BEVFusion/bevfusion/transformer.py#L14

Suggest to add BatchNorm to the positional embedding as well.

autoware_ml/models/detection3d/heads/transfusion.py
```diff
@@ -127,15 +127,26 @@ def forward(
...
+        )
         query_tokens = self.norm1(query_tokens + self.dropout(attended))
 
-        attended, _ = self.cross_attn(query_tokens, key_tokens, key_tokens)
+        attended, _ = self.cross_attn(
+            query_tokens + query_pos_embed,
+            key_tokens + key_pos_embed,
+            key_tokens + key_pos_embed,
+        )
         query_tokens = self.norm2(query_tokens + self.dropout(attended))
```
KSeangTan, line 151: Seems that the BEVFusion implementation doesn't have residual in query_tokens + attended in norm, might be a bug or a design choice 

https://github.com/tier4/AWML/blob/main/projects/BEVFusion/bevfusion/transformer.py#L86-L110

autoware_ml/preprocessing/detection3d/point_pillar.py
```diff
@@ -66,6 +69,7 @@ def __init__(
         )
         self.max_num_points = max_num_points
         self.max_voxels = max_voxels
+        self.eval_max_voxels = max_voxels if eval_max_voxels is None else eval_max_voxels
```
KSeangTan: We should raise an error when the mode is eval and eval_max_voxels doesn't set.

autoware_ml/preprocessing/base.py
```diff
@@ -55,6 +57,26 @@ def __init__(self, pipeline: Sequence[Any] = ()) -> None:
...
+        Args:
+            mode: ``True`` for training behavior, ``False`` for evaluation.
+
+        Returns:
+            This ``DataPreprocessing`` instance for chaining.
+        """
+        for layer in self.pipeline:
+            if isinstance(layer, nn.Module):
+                layer.train(mode)
+        return self
```
KSeangTan: Do we need to return a `self` here?

autoware_ml/models/base.py
```diff
@@ -104,6 +104,9 @@ def on_after_batch_transfer(
             Batch dictionary after runtime preprocessing.
         """
         del dataloader_idx
+        # The preprocessing pipeline is not a registered submodule, so keep its
+        # train/eval mode in sync with the model before applying it.
+        self._data_preprocessing.train(self.training)
```
KSeangTan: self._data_preprocessing.train(self.training) should be called only in:
```
on_test_start
on_validation_start
on_predict_start
```
Switch back to the train mode in:
```
on_test_end
on_validation_end
on_predict_end
```

Another cleaner alternative is to pass `is_training` to `__call__` in `preprocessing`. If any preprocessing module needs to differentiate between training/eval, they should utilize the parameter.

autoware_ml/transforms/point_cloud/sweeps.py
```diff
@@ -71,6 +78,9 @@ def __init__(
         self.pad_empty_sweeps = pad_empty_sweeps
         self.remove_close = remove_close
         self.close_radius = close_radius
+        if selection not in {"nearest", "random"}:
```
KSeangTan: `test_mode` is a cleaner and better alternative. We should always assume selection is random in training, and use only `nearest` in testing.

KSeangTan (changes requested):
Please run unit tests and make sure each unit test pass. We will automate the process eventually, for now, please run them manually. If there's no unit tests, please add fews of them.

autoware_ml/preprocessing/base.py
```diff
@@ -55,6 +57,26 @@ def __init__(self, pipeline: Sequence[Any] = ()) -> None:
...
+
```
amadeuszsz: Can we remove this change? Instead:
- remove `nn.Module` inheritance from `class PointPillarPreprocessor`
- replace `forward` with `__call__` in `class PointPillarPreprocessor`
- replace `register_buffer` calls with plain `self.voxel_size = torch.tensor(...)` in `class PointPillarPreprocessor`

autoware_ml/models/detection3d/heads/transfusion.py
vividf: The residual is actually there in the reference — it is just hidden inside mmcv's `MultiheadAttention` wrapper, so it doesn't show up in the BEVFusion-side code you linked.

The reference layer calls `self.self_attn(query=query, ..., value=query + query_pos)` without passing `identity`, so mmcv falls back to `identity = query` internally, and the wrapper returns `identity + ...` at the end. From mmcv v2.1.0 (the version pinned in the AWML Dockerfile), [`mmcv/cnn/bricks/transformer.py#L516-L552`](https://github.com/open-mmlab/mmcv/blob/v2.1.0/mmcv/cnn/bricks/transformer.py#L516-L552):

```python
if identity is None:
    identity = query          # saved BEFORE query_pos is added
...
if query_pos is not None:
    query = query + query_pos # pos enc only enters the attention inputs
if key_pos is not None:
    key = key + key_pos
...
out = self.attn(query=query, key=key, value=value, ...)[0]
...
return identity + self.dropout_layer(self.proj_drop(out))   # <-- the residual
```

The FFN wrapper does the same thing ([`transformer.py#L629-L635`](https://github.com/open-mmlab/mmcv/blob/v2.1.0/mmcv/cnn/bricks/transformer.py#L629-L635)): `return identity + self.dropout_layer(out)`.

Two things follow from this:

1. The reference does have a residual on every attention/FFN block — it's just written inside the wrapper instead of at the call site. Since this repo uses PyTorch's native `nn.MultiheadAttention` (which does not add a residual internally), the residual has to be written out explicitly here, which is what `self.norm1(query_tokens + self.dropout(attended))` does. The two are mathematically identical.
2. Note that `identity` is captured *before* `query_pos` is added, i.e. the residual stream in the reference stays position-free and the positional embedding is re-added at each attention call. That is exactly the behavior this PR aligns to.

So no code change needed here — the implementation already matches the reference.

autoware_ml/models/detection3d/heads/transfusion.py
vividf: The mmdet code you linked is not what actually runs — BEVFusion's `TransformerDecoderLayer` builds its attention from mmcv's `MultiheadAttention` wrapper, and that wrapper *does* apply dropout to the attended values on its way back into the residual, in mmcv v2.1.0 (the version pinned in AWML), [`mmcv/cnn/bricks/transformer.py#L552`](https://github.com/open-mmlab/mmcv/blob/v2.1.0/mmcv/cnn/bricks/transformer.py#L552):

```python
return identity + self.dropout_layer(self.proj_drop(out))
```

The key detail is how the deprecated `dropout` kwarg is handled in `__init__` ([`transformer.py#L440-L447`](https://github.com/open-mmlab/mmcv/blob/v2.1.0/mmcv/cnn/bricks/transformer.py#L440-L447)):

```python
if 'dropout' in kwargs:
    warnings.warn(...)
    attn_drop = kwargs['dropout']
    dropout_layer['drop_prob'] = kwargs.pop('dropout')
```

and the AWML reference config uses exactly that deprecated kwarg ([`default_lidar_second_secfpn_120m.py#L58-L59`](https://github.com/tier4/AWML/blob/main/projects/BEVFusion/configs/t4dataset/default/models/default_lidar_second_secfpn_120m.py#L58-L59)):

```python
self_attn_cfg=dict(embed_dims=128, num_heads=8, dropout=0.1),
cross_attn_cfg=dict(embed_dims=128, num_heads=8, dropout=0.1),
```

So what the reference actually runs is `attn_drop=0.1` **and** an output-side dropout of 0.1 before the residual add (with `proj_drop=0`). Our `nn.MultiheadAttention(dropout=dropout)` + `self.norm1(query_tokens + self.dropout(attended))` is the same thing: attention-weight dropout 0.1 + output dropout 0.1 before the residual.

The FFN side is also equivalent: the reference `FFN(ffn_drop=0.1)` runs `Linear → ReLU → Dropout → Linear → Dropout` and then adds identity (its `dropout_layer` is not configured → Identity), while ours runs `Linear → ReLU → Dropout → Linear` followed by the external `self.dropout` before the residual — same positions, same probabilities.

On splitting into separate dropout knobs: happy to do it, but note it wouldn't be three values — mmcv's granularity is actually four (`attn_drop`, `proj_drop`, `dropout_layer`, `ffn_drop`). Since every config we have uses a single 0.1 for all of them (matching the reference), I'd prefer to keep the single `dropout` argument in this PR and split it if/when a config actually needs different values.

autoware_ml/configs/tasks/detection3d/bevfusion/camera_lidar_voxel0075_second_secfpn_54m_nuscenes.yaml
vividf: Renamed `selection` to `test_mode` in ecbaf6f53edd2b5b4c25882856231671769f5237, matching the existing `LoadMultiSweepPointsFromFile` convention.

autoware_ml/configs/tasks/detection3d/bevfusion/camera_lidar_voxel0170_second_secfpn_120m_t4dataset_j6gen2.yaml
vividf: Fixed in ecbaf6f53edd2b5b4c25882856231671769f5237.

autoware_ml/transforms/point_cloud/sweeps.py
vividf: Agreed — switched to `test_mode` in ecbaf6f53edd2b5b4c25882856231671769f5237: training always samples randomly, testing always takes the nearest sweeps.

autoware_ml/models/common/layers/conv.py
vividf: Thanks, addressed in 1a945508b4e70f4c9a5e486863559abc7e165812.

autoware_ml/models/detection3d/heads/transfusion.py
vividf: You are right — the BatchNorm1d was missing. Fixed in 25b433cd65a0429f44f999e2431c4ce229919d14: `LearnedPositionalEncoding` now mirrors the reference `PositionEncodingLearned` layer-for-layer (`Conv1d → BatchNorm1d → ReLU → Conv1d`, BN with PyTorch defaults as in the reference). Note this changes the module state_dict (Linear → Conv1d weights + BN buffers), so checkpoints trained before this commit need re-conversion — the upside is that reference checkpoints now convert directly without folding the BN into the linear layer.

autoware_ml/preprocessing/detection3d/point_pillar.py
vividf: Done in 09c7a5c — evaluation mode now raises when `eval_max_voxels` is not set. The centerpoint and transfusion configs set it explicitly (same value as `max_voxels`) to keep their previous behavior.

autoware_ml/models/base.py
vividf: Went with your cleaner alternative in 09c7a5c: `DataPreprocessing.__call__` now takes an explicit `is_training` argument and the per-batch `train()` sync is gone. I avoided the lifecycle-hook option because `deploy.py` calls `on_after_batch_transfer` directly without a Trainer loop, so `on_predict_start` would never fire there and export would silently run the preprocessor in training mode. Passing `is_training=self.training` covers that path too.

autoware_ml/preprocessing/base.py
vividf: The `train()` method is removed entirely in 09c7a5c, so this is moot now — the mode is passed as an explicit `is_training` call argument instead.

autoware_ml/preprocessing/base.py
vividf: Done in 09c7a5c — the `train()` propagation is removed and `PointPillarPreprocessor` no longer inherits `nn.Module`: `forward` is now `__call__`, the `register_buffer` calls are plain tensors, and the train/eval mode arrives as an explicit `is_training` argument passed by `DataPreprocessing`.

vividf:
Added unit tests in 09c7a5c covering the behaviors this PR changes:

- `remove_close` box-vs-circle semantics — (0.9, 0.9) is inside the box but outside the r=1.0 circle, so it discriminates the two
- sweep selection: nearest in test mode, uniform sampling without replacement in training
- train/eval voxel budgets, including the new error when `eval_max_voxels` is unset in evaluation
- `is_training` forwarding through `DataPreprocessing`

All affected suites pass locally (103 tests); `test_transfusion.py` needs the docker image (spconv) and will be verified there.

autoware_ml/datamodule/common/detection3d.py
vividf: These changes are from another stacked PR. Because of the GitHub issue, I will make another PR to handle this

amadeuszsz (approved):
Thank you for this PR!

I do not like the introduction of `is_training` argument, it pollutes design. At the same time I understand it was not possible to handle different number of voxels configuration without access to the trainer instance (so you can't get the runner state).

I think we may need to handle preprocessing pipeline per-split, but this is a change for a separate PR.

LGTM!

KSeangTan (approved):
LGTM overall, we should merge it before proceeding to other changes.

PR #89, opened 2026-08-07
feat(deploy): add deploy.onnx.precision option for explicit-fp16 exports

mojomex:
## Summary

Adds a `deploy.onnx.precision` option (`fp32` | `fp16`, default `fp32`) to the ONNX export stage.

- `fp32` exports the model exactly as before.
- `fp16` converts the export to half precision **internally, behind an fp32 graph boundary**: initializers and internal tensors become fp16, while graph inputs/outputs stay fp32 via cast layers inside the model. Engine I/O therefore keeps its dtype contract and consumers keep their fp32 buffers, whatever the export precision.

The exported ONNX is the single source of truth for an engine's numerics. The TensorRT stage builds strongly typed, as it always has, so the builder uses the dtypes the ONNX carries instead of assigning its own — picking a precision is a `deploy.onnx.precision` decision only, and there is deliberately no TensorRT-side typing knob.

Reduced-precision exports have to be built this way: the weakly typed precision-assignment pass would re-process the already-reduced tensors, and on one of our embedded targets it crashes on PTv3-shaped graphs regardless of builder settings. Consuming inference nodes must honor the export's dtypes too — in Autoware that is `trt_precision: strongly-typed` (autoware_universe counterpart PRs).

The conversion (in the new `autoware_ml/utils/onnx_precision.py`, wired into `scripts/deploy.py` after graph modifiers) handles three pitfalls of `onnxconverter-common` on these graphs:

- graph outputs that double as internal inputs (PTv3's `point_feat_0` feeds the next encoder stage): internal consumers are rewired to the pre-cast fp16 tensor, since the fp32 output cast must not re-enter the fp16 interior;
- pre-existing no-op `Cast(to=FLOAT)` nodes (e.g. the post-softmax `.to(query.dtype)` pattern) are retargeted to fp16 — except the boundary casts producing graph outputs — since a strongly typed parse rejects the mixed-precision consumers they would otherwise create;
- the op block list is empty, because the custom plugin ops negotiate their own precisions and support fp16 I/O natively.

Precisions that require calibration (int8) are explicitly out of scope.

## Related Issues

- autoware_universe counterparts: autowarefoundation/autoware_universe#13160 (strongly typed builds in `autoware_tensorrt_common`, merged), autowarefoundation/autoware_universe#13191 (names the precision value `"strongly-typed"`), autowarefoundation/autoware_universe#13159 (PTv3 adoption)
- Related optimization (independent): autowarefoundation/autoware_universe#13158

## Checklist

## Additional Notes

- The option is documented in `docs/user-guide/deployment.md` rather than in comments in `deploy.yaml`, per review.
- `deploy.onnx.precision` is typed as an `OnnxPrecision(StrEnum)` with `auto()` values, per review, so the accepted spellings and the config strings are the same thing.
- `pixi.lock` is regenerated with the repo-pinned pixi 0.66.0 (in docker, keeping the v6 format): `onnxconverter-common==1.16.0` is locked in the `default` and `dev` environments (+17 lines, nothing else re-solved). The dependency is required — it is not pulled in transitively.
- Models containing subgraphs (`If`/`Loop`/`Scan`) are refused with an explicit `NotImplementedError`: the cast fixups only handle the top-level graph, and no autoware-ml model uses subgraphs.
- Unit tests cover the precision option validation and the conversion pitfalls — including the dual-use-output rewiring and the fp32 boundary contract — on a minimal synthetic graph (`tests/utils/test_onnx_precision.py`). Full suite in the project image: **507 passed**, against 502 at the base commit — exactly the four new tests, nothing else moved. (The 2 collection errors in both runs are the `bev_pool` CUDA extension not being built in that workspace.)
- Note for whoever rebuilds the published image: `onnxconverter-common` is new in the lockfile, and `scripts/deploy.py` imports it unconditionally, so a container built from the previous lock fails on every deploy — including the default fp32 path — until the image is rebuilt.
- fp16 accuracy validation against the fp32 reference is pending; this PR is about producing deployable builds.

## Additional Log and Media

Verified with the PTv3 HuggingFace package: the converted encoder, seg3d head and det3d head keep fp32 graph I/O (`io dtypes = FLOAT/INT32/INT64`, 210/86/157 fp16 initializers) and build as strongly typed TensorRT engines through the unmodified PTv3 node buffers, on x86 and on the embedded target, where the weakly typed fp16 build crashes.

### DIFF
--- a/autoware_ml/configs/defaults/modules/deploy.yaml
+++ b/autoware_ml/configs/defaults/modules/deploy.yaml
@@ -6,6 +6,7 @@ deploy:
     dynamo: true
     opset_version: 21
     modify_graph: null
+    precision: fp32
   tensorrt:
     enabled: true
     workspace_size: 4294967296
--- a/autoware_ml/tests/utils/test_onnx_precision.py
+++ b/autoware_ml/tests/utils/test_onnx_precision.py
@@ -1,0 +1,136 @@
+# Copyright 2026 TIER IV, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Tests for export-time ONNX precision conversion."""
+
+from __future__ import annotations
+
+import numpy as np
+import onnx
+import pytest
+from omegaconf import OmegaConf
+from onnx import TensorProto, helper, numpy_helper
+
+from autoware_ml.utils.onnx_precision import (
+    OnnxPrecision,
+    convert_onnx_precision,
+    resolve_onnx_precision,
+    should_convert_precision,
+)
+
+
+def _make_fp32_model() -> onnx.ModelProto:
+    """A minimal fp32 graph exercising the conversion pitfalls.
+
+    MatMul with an fp32 weight initializer, followed by a pre-existing no-op Cast(to=FLOAT)
+    (the pattern torch exports for ``.to(query.dtype)``), whose output is both a graph output
+    and an internal input to a second MatMul — mirroring PTv3's dual-use point_feat_0.
+    """
+    weight = numpy_helper.from_array(np.eye(4, dtype=np.float32), name="weight")
+    nodes = [
+        helper.make_node("MatMul", ["x", "weight"], ["mm0"], name="mm0"),
+        helper.make_node("Cast", ["mm0"], ["feat"], name="noop_cast", to=TensorProto.FLOAT),
+        helper.make_node("MatMul", ["feat", "weight"], ["out"], name="mm1"),
+    ]
+    graph = helper.make_graph(
+        nodes,
+        "tiny",
+        inputs=[helper.make_tensor_value_info("x", TensorProto.FLOAT, [None, 4])],
+        outputs=[
+            helper.make_tensor_value_info("feat", TensorProto.FLOAT, [None, 4]),
+            helper.make_tensor_value_info("out", TensorProto.FLOAT, [None, 4]),
+        ],
+        initializer=[weight],
+    )
+    return helper.make_model(graph, opset_imports=[helper.make_opsetid("", 17)])
+
+
+def test_resolve_precision_defaults_to_fp32() -> None:
+    assert resolve_onnx_precision(OmegaConf.create({})) is OnnxPrecision.FP32
+    assert not should_convert_precision(OmegaConf.create({}))
+
+
+def test_resolve_precision_reads_config_string() -> None:
+    """The enum's values are the spellings used in deploy.onnx.precision."""
+    cfg = OmegaConf.create({"precision": "fp16"})
+    assert resolve_onnx_precision(cfg) is OnnxPrecision.FP16
+    assert should_convert_precision(cfg)
+
+
+def test_resolve_precision_rejects_unsupported() -> None:
+    with pytest.raises(ValueError, match="int8"):
+        resolve_onnx_precision(OmegaConf.create({"precision": "int8"}))
+
+
+def test_fp16_conversion(tmp_path) -> None:
+    onnx_path = tmp_path / "tiny.onnx"
+    onnx.save(_make_fp32_model(), onnx_path.as_posix())
+
+    result = convert_onnx_precision(onnx_path, OnnxPrecision.FP16)
+    model = onnx.load(result.as_posix())
+
+    initializer_types = {i.name: i.data_type for i in model.graph.initializer}
+    assert initializer_types["weight"] == TensorProto.FLOAT16
+
+    # The graph boundary stays fp32: reduced precision is internal, behind cast layers, so
+    # consumers keep their fp32 buffers.
+    for value_info in list(model.graph.input) + list(model.graph.output):
+        assert value_info.type.tensor_type.elem_type == TensorProto.FLOAT
+
+    # 'feat' is both a graph output and an internal input of the second MatMul. The output must
+    # be produced by an fp16->fp32 cast, while the internal consumer must read the pre-cast fp16
+    # tensor - otherwise a strongly typed parse rejects the mixed Float/Half MatMul.
+    producers = {output: node for node in model.graph.node for output in node.output}
+    feat_producer = producers["feat"]
+    assert feat_producer.op_type == "Cast"
+    assert next(a.i for a in feat_producer.attribute if a.name == "to") == TensorProto.FLOAT
+    internal_consumers = [
+        n for n in model.graph.node if "feat" in n.input and n is not feat_producer
+    ]
+    assert not internal_consumers, "internal consumers must be rewired to the fp16 tensor"
+    mm1 = next(n for n in model.graph.node if n.name == "mm1")
+    assert mm1.input[0] == feat_producer.input[0]
+
+    onnx.checker.check_model(model)
+
+
+def test_fp16_conversion_rejects_subgraphs(tmp_path) -> None:
+    """Models with If/Loop/Scan subgraphs are refused: the cast fixups only handle the top level."""
+    const = helper.make_node(
+        "Constant",
+        [],
+        ["y"],
+        value=helper.make_tensor("v", TensorProto.FLOAT, [1], [1.0]),
+    )
+    branch = helper.make_graph(
+        [const],
+        "branch",
+        inputs=[],
+        outputs=[helper.make_tensor_value_info("y", TensorProto.FLOAT, [1])],
+    )
+    if_node = helper.make_node(
+        "If", ["cond"], ["y"], name="if0", then_branch=branch, else_branch=branch
+    )
+    graph = helper.make_graph(
+        [if_node],
+        "with_subgraph",
+        inputs=[helper.make_tensor_value_info("cond", TensorProto.BOOL, [])],
+        outputs=[helper.make_tensor_value_info("y", TensorProto.FLOAT, [1])],
+    )
+    model = helper.make_model(graph, opset_imports=[helper.make_opsetid("", 17)])
+    onnx_path = tmp_path / "subgraph.onnx"
+    onnx.save(model, onnx_path.as_posix())
+
+    with pytest.raises(NotImplementedError, match="if0"):
+        convert_onnx_precision(onnx_path, OnnxPrecision.FP16)
--- a/autoware_ml/utils/deploy.py
+++ b/autoware_ml/utils/deploy.py
@@ -431,6 +431,8 @@ def create_tensorrt_builder_config(tensorrt_cfg: DictConfig) -> tuple[Any, Any, 
     trt_logger = trt.Logger(trt.Logger.WARNING)
     trt.init_libnvinfer_plugins(trt_logger, "")
     builder = trt.Builder(trt_logger)
+    # Always strongly typed: deploy.onnx.precision decides which dtypes the ONNX carries, and the
+    # engine has to use them as exported rather than let the builder reassign precisions.
     network = builder.create_network(1 << int(trt.NetworkDefinitionCreationFlag.STRONGLY_TYPED))
     parser = trt.OnnxParser(network, trt_logger)
     config = builder.create_builder_config()
--- a/autoware_ml/utils/onnx_precision.py
+++ b/autoware_ml/utils/onnx_precision.py
@@ -1,0 +1,175 @@
+# Copyright 2026 TIER IV, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Explicit-precision conversion for exported ONNX models.
+
+``deploy.onnx.precision`` controls the float dtypes baked into the exported ONNX.
+The TensorRT stage always builds strongly typed, so those dtypes are what the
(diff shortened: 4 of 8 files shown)


autoware_ml/configs/defaults/modules/deploy.yaml
```diff
@@ -6,6 +6,21 @@ deploy:
...
+    # export to explicit half precision. Precisions requiring calibration (int8, ...) are not
+    # supported here.
+    #
+    # Precision and strong typing are separate concepts: precision is a property of the ONNX,
+    # strong typing a property of the TensorRT build (see deploy.tensorrt.strongly_typed and the
+    # ml_package strongly_typed flag consumed by the inference nodes). Any combination is valid.
+    # Reduced-precision exports should be built strongly typed: the weakly typed builder would
+    # re-run its precision assignment on the already-reduced tensors (and crashes doing so on
+    # DriveOS TensorRT 10.10), so ship them with `strongly_typed: true` in each model's
+    # ml_package param file.
```
mojomex:
```suggestion
    # Precision of the exported ONNX artifact: fp32 exports the model as-is, fp16 converts the
    # export to explicit half precision. Precisions requiring calibration (int8, ...) are not
    # supported here.
```

autoware_ml/configs/defaults/modules/deploy.yaml
```diff
@@ -6,6 +6,21 @@ deploy:
...
+    precision: fp32
   tensorrt:
     enabled: true
     workspace_size: 4294967296
+    # Build the validation engine strongly typed (tensor precisions fixed by the ONNX) rather
+    # than weakly typed (builder assigns precisions). Orthogonal to deploy.onnx.precision.
```
mojomex:
```suggestion
    # Declare the model as strongly typed (tensor precisions fixed by the ONNX) rather
    # than weakly typed (TensorRT builder overrides precisions). Orthogonal to deploy.onnx.precision.
```

autoware_ml/utils/deploy.py
```diff
@@ -431,9 +431,17 @@ def create_tensorrt_builder_config(tensorrt_cfg: DictConfig) -> tuple[Any, Any,
     trt_logger = trt.Logger(trt.Logger.WARNING)
     trt.init_libnvinfer_plugins(trt_logger, "")
     builder = trt.Builder(trt_logger)
-    network = builder.create_network(1 << int(trt.NetworkDefinitionCreationFlag.STRONGLY_TYPED))
+    # Strong typing is orthogonal to the exported precision: any model (fp32 or reduced
+    # precision) can be built strongly typed, fixing every tensor precision to what the ONNX
+    # declares instead of letting the builder assign precisions.
```
mojomex:
```suggestion
    # Strong typing is orthogonal to the exported precision: any model (fp32 or reduced
    # precision) can be declared strongly typed, fixing every tensor precision to what the ONNX
    # declares instead of letting the TensorRT builder override precisions.
```

autoware_ml/utils/onnx_precision.py
```diff
@@ -0,0 +1,157 @@
...
+def _convert_to_fp16(onnx_path: Path) -> Path:
+    import onnx
+
+    try:
+        from onnxconverter_common import float16
+    except ImportError as error:  # pragma: no cover - environment dependent
+        raise RuntimeError(
+            "deploy.onnx.precision=fp16 requires the 'onnxconverter-common' package. "
+            "Install it in the deployment environment."
+        ) from error
```
mojomex: No conditional nor non-top-level imports. These are covered by pyproject.toml, so definitely exist.

autoware_ml/tests/utils/test_onnx_precision.py
```diff
@@ -0,0 +1,99 @@
...
+from omegaconf import OmegaConf
+from onnx import TensorProto, helper, numpy_helper
+
+from autoware_ml.utils.onnx_precision import (
+    convert_onnx_precision,
+    resolve_onnx_precision,
+    should_convert_precision,
+)
+
+pytest.importorskip("onnxconverter_common")
```
mojomex: This 100% exists since covered by pyproject.toml.

autoware_ml/utils/onnx_precision.py
```diff
@@ -0,0 +1,157 @@
...
+``deploy.onnx.precision`` controls the dtypes baked into the exported ONNX; whether TensorRT
+honors them verbatim is the separate strong-typing concept (a build property: any export can be
+built strongly typed, fixing tensor precisions to the model instead of leaving them to the
+builder). Reduced-precision exports should be built strongly typed — the weakly typed precision
+machinery would re-process the already-reduced tensors — so ship them with
+``strongly_typed: true`` in each model's ml_package param file, which the inference nodes consume.
+
+The graph boundary always stays fp32: reduced precision is internal, entered and left through
+cast layers inside the model. Engine I/O therefore keeps its dtype contract and consumers keep
+their fp32 buffers, whatever the export precision.
```
mojomex:
```suggestion
``deploy.onnx.precision`` controls the float dtypes baked into the exported ONNX.
This is separate from ``deploy.tensorrt.strongly_typed``, which toggles whether
the TensorRT builder shall honor the ONNX's types (`true`) or is free to override
types (`false`), e.g. build an `fp16` engine from an `fp32` ONNX.

All float network inputs/outputs always stay FP32, so inference code works
with all model precisions without interface changes.
```

autoware_ml/utils/onnx_precision.py
```diff
@@ -0,0 +1,157 @@
...
+    # casts the converter places incorrectly around ops it cannot shape-infer.
+    converted = float16.convert_float_to_float16(
+        model,
+        keep_io_types=True,
+        op_block_list=[],
+        disable_shape_infer=True,  # custom plugin ops break ONNX shape inference
+    )
+
+    rewired = _rewire_dual_use_outputs(onnx, converted)
+    recast = _retarget_stray_float_casts(onnx, converted)
```
mojomex:
```suggestion
    n_rewired = _rewire_dual_use_outputs(onnx, converted)
    n_recast = _retarget_stray_float_casts(onnx, converted)
```

autoware_ml/utils/onnx_precision.py
```diff
@@ -0,0 +1,157 @@
...
+    logger.info(
+        "Converted %s to fp16 with fp32 graph I/O (rewired %d dual-use output consumer(s), "
+        "retargeted %d pre-existing float cast(s))",
+        onnx_path,
+        rewired,
+        recast,
```
mojomex:
```suggestion
        n_rewired,
        n_recast,
```

autoware_ml/utils/onnx_precision.py
```diff
@@ -0,0 +1,157 @@
...
+    )
+
+    onnx.save(converted, onnx_path.as_posix())
+    return onnx_path
+
+
+def _rewire_dual_use_outputs(onnx: Any, model: Any) -> int:
+    """Rewire internal consumers of fp32 graph outputs to the pre-cast fp16 tensor.
+
+    A graph output can double as an internal input (PTv3's point_feat_0 feeds the next encoder
```
mojomex:
```suggestion
    A graph output can double as an internal input (e.g. PTv3's point_feat_0 feeds the next encoder
```

autoware_ml/utils/onnx_precision.py
```diff
@@ -0,0 +1,157 @@
...
+    stage). keep_io_types casts such outputs back to fp32, which would re-enter the fp16 interior
+    as fp32 and fail a strongly typed parse; internal consumers must read the fp16 tensor the
+    output cast was fed from instead.
+    """
+    graph_outputs = {output.name for output in model.graph.output}
+    producers = {output: node for node in model.graph.node for output in node.output}
+    rewired = 0
```
mojomex:
```suggestion
    n_rewired = 0
```

autoware_ml/utils/onnx_precision.py
```diff
@@ -0,0 +1,157 @@
...
+        producer = producers.get(name)
+        if producer is None or producer.op_type != "Cast":
+            continue
+        cast_target = next((a.i for a in producer.attribute if a.name == "to"), None)
+        if cast_target != onnx.TensorProto.FLOAT:
+            continue
+        pre_cast_tensor = producer.input[0]
+        for node in model.graph.node:
+            if node is producer:
+                continue
```
mojomex: Rewrite this so we get an easily readable `consumers = ` expression

autoware_ml/utils/onnx_precision.py
```diff
@@ -0,0 +1,157 @@
...
+
+
+def _retarget_stray_float_casts(onnx: Any, model: Any) -> int:
+    """Retarget pre-existing ``Cast(to=FLOAT)`` nodes (and float ConstantOfShape values) to fp16.
+
+    The converter rewrites tensors and initializers but leaves the ``to`` attribute of original
+    Cast nodes untouched. Exports commonly contain fp32->fp32 no-op casts (e.g. the post-softmax
+    ``.to(query.dtype)`` in attention); after conversion these would inject fp32 into fp16
+    consumers, which a strongly typed parse rejects. Casts that produce a graph output are the
+    fp32 boundary and stay untouched.
```
mojomex:
```suggestion
    ``float16.convert_float_to_float16`` leaves the ``to`` attribute of
    Cast nodes untouched. Exports commonly contain fp32->fp32 no-op casts (e.g. the post-softmax
    ``.to(query.dtype)`` in attention). After conversion these would cast fp16->fp32, feeding the wrong precision to consumers.
    Casts that produce a graph output are the
    fp32 boundary and stay untouched.
```

autoware_ml/utils/onnx_precision.py
```diff
@@ -0,0 +1,157 @@
...
+    """
+    graph_outputs = {output.name for output in model.graph.output}
+    recast = 0
```
mojomex:
```suggestion
    n_recast = 0
```

autoware_ml/configs/defaults/modules/deploy.yaml
```diff
@@ -6,6 +6,13 @@ deploy:
...
     modify_graph: null
+    # Precision of the exported ONNX artifact: fp32 exports the model as-is, fp16 converts the
+    # export to explicit half precision. Precisions requiring calibration (int8, ...) are not
+    # supported here.
+    precision: fp32
   tensorrt:
     enabled: true
     workspace_size: 4294967296
+    # Declare the model as strongly typed (tensor precisions fixed by the ONNX) rather
+    # than weakly typed (TensorRT builder overrides precisions). Orthogonal to deploy.onnx.precision.
```
mojomex:
```suggestion
    # Declare the model as strongly typed (TensorRT has to use tensor precisions defined ONNX) rather
    # than weakly typed (TensorRT builder is allowed to override precisions). Orthogonal to deploy.onnx.precision.
```

mojomex: Applied in 20f9950 — with one tweak: "precisions defined ONNX" read like a typo, so it says "defined in the ONNX" (and the lines are re-wrapped for yamllint). Shout if the original was intentional.

amadeuszsz:
@codex review

autoware_ml/utils/onnx_precision.py
```diff
@@ -0,0 +1,149 @@
...
+            for attribute in node.attribute:
+                if attribute.name == "to" and attribute.i == onnx.TensorProto.FLOAT:
+                    attribute.i = onnx.TensorProto.FLOAT16
+                    n_recast += 1
+        elif node.op_type == "ConstantOfShape":
+            for attribute in node.attribute:
+                if attribute.name == "value" and attribute.t.data_type == onnx.TensorProto.FLOAT:
+                    value = onnx.numpy_helper.to_array(attribute.t)
+                    attribute.t.CopyFrom(onnx.numpy_helper.from_array(value.astype("float16")))
+    return n_recast
```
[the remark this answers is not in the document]
mojomex: Per offline discussion: rather than making the conversion subgraph-aware, `_convert_to_fp16` now refuses models containing `GRAPH`/`GRAPHS` attributes up front with an explicit `NotImplementedError` naming the offending nodes (63ec40e). No autoware-ml model uses `If`/`Loop`/`Scan`, so recursive handling would be untested complexity; the guard turns the silent-miscompilation risk into a loud failure at export time. Covered by `test_fp16_conversion_rejects_subgraphs`.

pyproject.toml
```diff
@@ -28,6 +28,7 @@ dependencies = [
     "optuna==4.8.0",
     "onnx==1.21.0",
     "onnx_graphsurgeon==0.5.8",
+    "onnxconverter-common==1.16.0",
```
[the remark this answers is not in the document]
mojomex: Regenerated in 63ec40e with the repo-pinned pixi 0.66.0 in a fresh docker container, so the lock stays v6-format. The diff is +17 lines: `onnxconverter-common==1.16.0` added to the `default` and `dev` environments, nothing else re-solved. `pixi install --locked` flows accept the manifest again.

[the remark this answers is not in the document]
mojomex: Fixed in 63ec40e — lockfile regenerated with the pinned pixi 0.66.0 (docker), v6 format preserved; `onnxconverter-common==1.16.0` is now locked in both environments.

autoware_ml/configs/defaults/modules/deploy.yaml
```diff
@@ -6,6 +6,14 @@ deploy:
     dynamo: true
     opset_version: 21
     modify_graph: null
+    # Precision of the exported ONNX artifact: fp32 exports the model as-is, fp16 converts the
```
amadeuszsz: I'm not a fan of explaining the things in the config. You can give a hint (e.g. `ax = [0, 1]  # x, y`) or explain decision (e.g. `ax = [0, 1]  # ignore z after BEV projection`), but documenting variable/argument should be placed in the documentation.

Please consider if we can update documentation that it covers your new feature.

autoware_ml/utils/onnx_precision.py
```diff
@@ -0,0 +1,163 @@
...
+import logging
+from pathlib import Path
+from typing import Any
+
+import onnx
+from onnxconverter_common import float16
+
+logger = logging.getLogger(__name__)
+
+SUPPORTED_PRECISIONS = ("fp32", "fp16")
```
amadeuszsz: Can we use `StrEnum` and `auto()`?

autoware_ml/configs/defaults/modules/deploy.yaml
mojomex: Agreed — moved to `docs/user-guide/deployment.md` (f424eab)

autoware_ml/utils/onnx_precision.py
mojomex: Done in de8204e

amadeuszsz (approved):
LGTM

PR #93, opened 2026-08-25
fix(ptv3): pad last attention window by wrapping, not by repeating

mojomex:
Serialized attention works on fixed-size windows, so the last window of a sample is short whenever the voxel count is not a multiple of the patch size.

**Training** fills those slots by borrowing the tail of the preceding window, which makes the last window a full-size *sliding* window over the final `patch_size` tokens. Every slot holds a real token, so nothing needs masking.

**Export** instead used `clamp(arange(padded_n), max=n-1)` — repeats of the last serialized token, unmasked in the softmax. Every query in that window then sees one key many times over, and it dominates the attention.

## What each scheme puts in the last window

Two regimes, because training behaves differently in each. Indices below are positions in serialization order; every window except the last is full and identical under all three schemes.

```
CASE A -- n = 20 tokens, window K = 16  ->  padded to 32 tokens (two full windows)

Last window:
                      ┌ 4 real leftovers ┐┌───────── 12 slots to fill ─────────┐
  training             16 17 18 19          4  5  6  7  8  9 10 11 12 13 14 15
  inference (old)      16 17 18 19         19 19 19 19 19 19 19 19 19 19 19 19
  proposed             16 17 18 19          4  5  6  7  8  9 10 11 12 13 14 15   ← same as training

  training / proposed : borrow the tail of window 0, so the last window becomes a full
                        sliding window over the final 16 real tokens. All keys distinct.
  inference (old)     : token 19 fills every slot, so it appears 13x inside one softmax
                        and dominates all 16 queries.


CASE B -- n = 5 tokens, window K = 16  ->  one window, no previous window to borrow from

Last window:

  training              0  1  2  3  4                                      ← no padding at all
  inference (old)       0  1  2  3  4  4  4  4  4  4  4  4  4  4  4  4
  proposed              0  1  2  3  4  4  0  1  2  3  4  0  1  2  3  4

  training            : flash-attention's varlen path takes the true sequence length, so
                        there is no window to fill and nothing to mask.
  inference (old)     : key 4 appears 12x against 1x for keys 0-3       -> skew 12:1
  proposed            : keys cycle back through the sample, so 4 appears 4x
                        and 0-3 appear 3x each                          -> skew  4:3
```

Case A is the one that matters in practice: it fires on every frame whose voxel count is not a multiple of the window size. Case B is the rare one, and the only place the proposed scheme does something training does not.

## Measurement method

Both tables below report the **relative element-wise L2 norm** of the difference in the attention sublayer's output features, `‖out − ref‖₂ / ‖ref‖₂`, over the tensor stated in the table. The reference differs between the two tables and is named in each.

**"Affected tokens" means output rows, not inputs.** The input features are byte-identical between the two runs; what changes is the set of keys each query's window exposes it to. The affected rows are the `n mod K` tokens that land in the last window — last in *serialization* order, which is **not** the last rows of the tensor: under a scrambled order their storage indices are scattered through it. Every other output row is bit-identical.

Numbers are the mean over 10 random weight initialisations, with the serialization order held fixed within each run (self-attention is permutation-equivariant over its window, so leaving the order free contaminates the comparison). They are magnitudes on random weights, not properties of a trained model — the structure of the tables is the claim, not the decimals.

## This was never limited to small samples

**Reference: the training fill.** Case A above, across voxel counts:

| n | K | affected output rows (`n mod K`) | share of rows | rel. err on affected rows | rel. err over all rows |
| --- | --- | --- | --- | --- | --- |
| 17 | 16 | 1 | 5.88% | 2.20 | 0.5197 |
| 20 | 16 | 4 | 20.00% | 1.67 | 0.7224 |
| 24 | 16 | 8 | 33.33% | 1.24 | 0.7537 |
| 31 | 16 | 15 | 48.39% | 0.20 | 0.1352 |
| 32 | 16 | 0 | 0.00% | exact | 0.0000 |
| 10000 | 256 | 16 | 0.16% | 2.47 | 0.0985 |

Exact multiples of the window size agree perfectly; **everything else diverged**. Two effects move in opposite directions, so read the last two columns together:

- **Severity** tracks the pad-slot count `K − r`. At `r = 1` a single key is duplicated 15 times and the affected row is off by 2.2× its own magnitude. At `r = 15` there is one duplicate and the error is 0.20.
- **Prevalence** tracks `r / n`, which shrinks as the sample grows.

Their product is why the all-rows column peaks in the middle rather than at either extreme, and why the last row matters most: at `n = 10000, K = 256` the all-rows figure collapses to 0.0985, but the 16 affected rows are still off by 2.47× their own magnitude. **Dilution is not mitigation** — at deployment scale a small, fixed set of output rows per frame was essentially garbage, which is exactly the kind of defect an aggregate metric hides.

## The fix

For slot `i` of the padded index array:

```
pad[i] = i                                  if i < n
       = (i - K + ceil(K / n) * n) mod n    otherwise
```

**For `n > K`** the added multiple of `n` is exactly `n`, the remainder is inert, and the expression reduces to `i - K` — the same backward borrow training performs. Verified identical to the training fill for `K` in {4, 16, 48, 1024} across every `n` up to 5·K, so Table 1 becomes all-zero by construction.

**For `n <= K`** there is no preceding window to borrow from, so it wraps around and cycles through the sample. Cycling keeps every key multiplicity within one of every other, against up to `K-1` for the old clamp, and is **exact whenever `n` divides `K`**: a uniform multiplicity cancels in the softmax, since `(K/n)·exp(sⱼ) / Σ(K/n)·exp(sᵢ)` reduces to `exp(sⱼ)/Σexp(sᵢ)`.

**Reference: exact attention over the `n` real tokens** — training does not pad at all in this regime, so there is no training fill to compare against. Here the whole sample is one window, so every output row is affected and the all-rows figure is the per-row figure:

| n | `n` divides `K` | rel. err, cyclic fill | rel. err, clamp fill |
| --- | --- | --- | --- |
| 2 | yes | **exact** | 0.8626 |
| 3 | no | 0.0756 | 1.0097 |
| 4 | yes | **exact** | 1.1833 |
| 5 | no | 0.0984 | 1.0672 |
| 8 | yes | **exact** | 0.9017 |
| 10 | no | 0.2007 | 0.6701 |
| 12 | no | 0.2570 | 0.5780 |
| 15 | no | 0.1431 | 0.1431 |
| 16 | yes | **exact** | **exact** |

With `K` a power of two, every power-of-two `n` is exact for free. The two schemes converge at `n = 15`, where the clamp has only one duplicate slot and coincides with the cycle.

The `ceil(K / n) * n` term keeps the shifted index non-negative, so the result does not depend on the runtime's sign convention for integer modulo — Python/numpy semantics and a C-style truncating `%` disagree on negative operands, and the latter would produce out-of-bounds gathers.

## No new data-dependent shapes

The window size stays static and `n` enters only as a value, so nothing in the graph gains a data-dependent shape. Shrinking the window for small samples would have introduced exactly that: the window extent is baked into `reshape(-1, patch_size, 3, H, D)`, and making it dynamic would put a runtime value in a reshape and require two code paths behind an `If` on a shape.

Because the export path is now total for any `n >= 1`, the guard requiring at least `patch_size` points per sample is removed, along with its `int(...item())` sync on a traced tensor — a constant-folding hazard as well as a false limit.

> [!IMPORTANT]
> This changes exported PTv3 numerics — **toward** what the model was trained on, so it is a correction rather than a drift, but checkpoints validated against the previous graph should be re-checked.

## Verification

In the `autoware-ml` container: **537 tests pass**.

Four regression tests added to `test_ptv3_encoder.py`, pinning each property above: the export fill equals the training fill for every `n % K != 0`, `pad` never leaves `[0, n)` for any n up to 4·K, the multiplicity spread stays within 1 below one window, and `n | K` is exact through the real module rather than a standalone reimplementation. The guard's test is replaced by one asserting a sub-window sample now exports with the window size unchanged.

## Outstanding

**TensorRT's integer `Mod` with a dynamic divisor is unverified.** The `ceil(K/n)*n` term removes the sign-convention dependency, but this has only been exercised in PyTorch and ONNX shape inference, not through an actual engine build. Worth a build on a real frame before relying on it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/autoware_ml/models/segmentation3d/encoders/ptv3.py
+++ b/autoware_ml/models/segmentation3d/encoders/ptv3.py
@@ -413,7 +413,7 @@ class SerializedAttention(PointModule):
             # this is exactly the backward borrow training performs; below one window
             # there is nothing to borrow from, so it wraps around instead. `cycle` keeps
             # the shifted index non-negative, so the modulo needs no particular sign
-            # convention from the runtime. See docs/models/ptv3.md for why it matters.
+            # convention from the runtime.
             divisor = torch.clamp(n, min=1)
             cycle = ((self.patch_size + divisor - 1) // divisor) * divisor
             index = torch.arange(padded_n, device=point.offset.device)
--- a/autoware_ml/models/segmentation3d/encoders/ptv3.py
+++ b/autoware_ml/models/segmentation3d/encoders/ptv3.py
@@ -409,19 +409,11 @@ class SerializedAttention(PointModule):
             padded_n = ((n + self.patch_size - 1) // self.patch_size) * self.patch_size
             unpad = torch.arange(n, device=point.offset.device)
             # Fill the trailing slots of the last window by cycling backwards through
-            # the serialization, so every slot holds a real token.
-            #
-            # For `n > patch_size` this borrows the tail of the preceding window and
-            # reproduces the training fill exactly: the last window becomes a full-size
-            # sliding window over the final `patch_size` tokens. With no preceding
-            # window to borrow from it wraps around instead, which keeps every key
-            # multiplicity within one of every other rather than repeating a single
-            # token to fill the gap - and is exact whenever `n` divides `patch_size`,
-            # because a uniform multiplicity cancels in the softmax.
-            #
-            # `cycle` is the smallest multiple of `n` that is at least `patch_size`, so
-            # the shifted index is never negative and the remainder needs no particular
-            # sign convention from the runtime's integer modulo.
+            # the serialization, so every slot holds a real token. For `n > patch_size`
+            # this is exactly the backward borrow training performs; below one window
+            # there is nothing to borrow from, so it wraps around instead. `cycle` keeps
+            # the shifted index non-negative, so the modulo needs no particular sign
+            # convention from the runtime. See docs/models/ptv3.md for why it matters.
             divisor = torch.clamp(n, min=1)
             cycle = ((self.patch_size + divisor - 1) // divisor) * divisor
             index = torch.arange(padded_n, device=point.offset.device)
--- a/autoware_ml/models/segmentation3d/encoders/ptv3.py
+++ b/autoware_ml/models/segmentation3d/encoders/ptv3.py
@@ -408,8 +408,24 @@ class SerializedAttention(PointModule):
             n = shape_as_tensor(point.feat)[0].to(device=point.offset.device)
             padded_n = ((n + self.patch_size - 1) // self.patch_size) * self.patch_size
             unpad = torch.arange(n, device=point.offset.device)
-            # Clamp pad indices to [0, n-1]: dummy slots repeat the last real token.
-            pad = torch.clamp(torch.arange(padded_n, device=point.offset.device), max=n - 1)
+            # Fill the trailing slots of the last window by cycling backwards through
+            # the serialization, so every slot holds a real token.
+            #
+            # For `n > patch_size` this borrows the tail of the preceding window and
+            # reproduces the training fill exactly: the last window becomes a full-size
+            # sliding window over the final `patch_size` tokens. With no preceding
+            # window to borrow from it wraps around instead, which keeps every key
+            # multiplicity within one of every other rather than repeating a single
+            # token to fill the gap - and is exact whenever `n` divides `patch_size`,
+            # because a uniform multiplicity cancels in the softmax.
+            #
+            # `cycle` is the smallest multiple of `n` that is at least `patch_size`, so
+            # the shifted index is never negative and the remainder needs no particular
+            # sign convention from the runtime's integer modulo.
+            divisor = torch.clamp(n, min=1)
+            cycle = ((self.patch_size + divisor - 1) // divisor) * divisor
+            index = torch.arange(padded_n, device=point.offset.device)
+            pad = torch.where(index < n, index, (index - self.patch_size + cycle) % divisor)
             if not self.enable_flash:
                 return pad, unpad, None
             # arange(0, padded_n+1, patch_size) gives [0, ps, 2*ps, ..., padded_n]
@@ -478,13 +494,14 @@ class SerializedAttention(PointModule):
         """
         head_count = self.num_heads
         if not self.enable_flash:
-            min_points = int(offset_to_bincount(point.offset).min().item())
-            if self.export_mode and min_points < self.patch_size_max:
-                raise ValueError(
-                    "PTv3 export mode requires each sample to have at least "
-                    f"{self.patch_size_max} serialized points, but found {min_points}."
-                )
-            self.patch_size = self.patch_size_max if self.export_mode else min_points
+            # Export keeps the window static so no shape depends on the voxel count;
+            # samples smaller than one window are handled by the cyclic fill in
+            # `_get_padding_and_inverse` rather than by shrinking the window.
+            self.patch_size = (
+                self.patch_size_max
+                if self.export_mode
+                else int(offset_to_bincount(point.offset).min().item())
+            )
         patch_size = self.patch_size
         channel_count = self.channels
         pad, unpad, cu_seqlens = self._get_padding_and_inverse(point)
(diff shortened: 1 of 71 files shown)


autoware_ml/tests/models/test_ptv3_encoder.py
```diff
@@ -524,3 +529,101 @@ def test_ptv3_frozen_encoder_supports_decoder_block_backward() -> None:
...
+        qk_scale=None,
+        attn_drop=0.0,
+        proj_drop=0.0,
+        order_index=0,
+        enable_rpe=False,
+        enable_flash=False,
+        upcast_attention=False,
+        upcast_softmax=False,
+    ).eval()
+    attention.patch_size = patch_size
```
mojomex, line 552: Is this needed?

docs/models/ptv3.md
```diff
@@ -133,6 +133,40 @@ graph contains no data-dependent pooling shape discovery. Pooled feature
...
+integer modulo.
+
+This matters beyond the small-sample case. Filling the trailing slots by
+repeating the last token instead - as a naive clamp would - leaves that token
+duplicated up to `K - 1` times inside an unmasked softmax, where it dominates
+every query in the window. That is a silent divergence from training on *every*
+frame whose voxel count is not a multiple of `patch_size`.
+
+Because the window size stays static and `n` enters only as a value, no shape in
+the graph depends on the data.
```
mojomex: Good but make much more concise.

autoware_ml/models/segmentation3d/encoders/ptv3.py
```diff
@@ -408,8 +408,24 @@ def _get_padding_and_inverse(
...
+            # reproduces the training fill exactly: the last window becomes a full-size
+            # sliding window over the final `patch_size` tokens. With no preceding
+            # window to borrow from it wraps around instead, which keeps every key
+            # multiplicity within one of every other rather than repeating a single
+            # token to fill the gap - and is exact whenever `n` divides `patch_size`,
+            # because a uniform multiplicity cancels in the softmax.
+            #
+            # `cycle` is the smallest multiple of `n` that is at least `patch_size`, so
+            # the shifted index is never negative and the remainder needs no particular
+            # sign convention from the runtime's integer modulo.
```
mojomex: Make much more concise.

autoware_ml/tests/models/test_ptv3_encoder.py
mojomex: Yes - `__init__` sets `self.patch_size = patch_size if enable_flash else 0`, and for the non-flash path `forward` is what assigns the real value. These tests call `_get_padding_and_inverse` directly, so without the reassignment the window size is 0. Added a two-line comment saying so rather than removing it.

docs/models/ptv3.md
mojomex: Cut from 31 prose lines to 16. Kept the index formula, the two regimes (`n > K` reduces to the training borrow, `n <= K` cycles), the sign-convention note, and why a clamp fill is wrong. Dropped the worked-out algebra and the separate "this matters beyond small samples" paragraph - the measurement tables in the PR description cover that.

autoware_ml/models/segmentation3d/encoders/ptv3.py
mojomex: Cut from 14 lines to 6 - what the fill does, the two regimes, and why `cycle` is there - with a pointer to `docs/models/ptv3.md` for the rationale so the two do not duplicate each other.

docs/models/ptv3.md
mojomex: Cut further in `00a5d7a` - the section is now three cases (exact / borrow from the preceding window / round-robin cycle) plus one line on why the fill exists at all, since dynamic window sizes and masking both cost performance in TensorRT. 12 lines, down from 33. The formula and the divergence measurements stay in the PR description.

amadeuszsz:
@codex review

amadeuszsz (approved):
LGTM!

PR #94, opened 2026-08-25
feat(litept): add LitePT as a gated PTv3 configuration

mojomex:
Adds [LitePT](https://arxiv.org/abs/2512.13689) seperate to, but based on, our existing PTv3 implementation.

## Changes compared to LitePT paper

I keep our PTv3 serialization, pooling and export contract exactly.
This means **LitePT is a fully compatible drop-in replacement** on the inference side.

The gated graph reads fewer of the declared inputs, so the ONNX exporter drops the ones it never traces and a LitePT artifact exposes a subset of the PTv3 inputs. The node binds what the artifact declares as of autowarefoundation/autoware_universe#13290, so a subset loads; that PR should land before a LitePT engine is deployed.

The original paper used GridPooling instead of PTv3's optimized SerializedPooling (mathematically the result is the same). It also computed serializations _inside_ the model, whereas we already do that in pre-processing to eliminate data-dependent shapes.

What's left: LitePT changes **which operations run at which resolution**: stages 0-2 are pure submanifold convolution, stages 3-4 pure attention with rotary position embedding (RoPE).

All existing PTv3 hyperparams apply 1:1 to LitePT.

## Implementation details

`LitePTEncoder` subclasses `PointTransformerV3Encoder`, so `PTv3SegmentationModel`, `PTv3DetectionModel`, `PTv3SegDetModel`, the decoder head and every export path accept it **unchanged**. The gating and the rotary embedding are options on the shared components, with defaults that leave PTv3 bit-for-bit as it was.

| | PTv3 | LitePT |
| --- | --- | --- |
| Block composition | every block: conv + attention + MLP | stages 0-2 conv only, 3-4 attention only |
| Positional encoding | submanifold conv per block | conv early; axis-split 3D RoPE on `grid_coord` late |
| Decoder | `dec_depths` blocks per stage | `base`: all zero, unpooling only; `tiny`: ptv3-tiny's blocks, gated |
| Encoder blocks | 14 conv + 14 attention | 6 conv + 8 attention (at `base`) |
| Encoder graph inputs | 27 | 20 (the exporter drops the 7 the gating never reads) |

## Configurations

LitePT mirrors ptv3's config structure, `base` and `tiny`, in `segmentation3d` and in `multi`. Each file carries only what differs from the ptv3 config it composes on top of.

`base` is the published LitePT parametrization: `enc_channels` 36/72/144/252/504 against 2/4/8/14/28 heads, and a decoder with no blocks that only unpools.

`tiny` is ptv3-tiny with the gating applied and nothing else changed, so the two are directly comparable. It keeps ptv3-tiny's decoder blocks and runs them under LitePT's rule - attention with rotary embedding, no submanifold convolution. The dataset configs are at `tiny` scale, following the ptv3 configs they inherit from.

## Rotary embedding and channel widths

The rotary embedding splits the head dimension into three per-axis chunks that each rotate in pairs, so it needs a multiple of 6; anything left over passes through unrotated (NoPE).

`base` follows the paper here: 18 channels per head at every stage, fully rotated. `tiny` inherits ptv3-tiny's multiple-of-32 widths, because many CUTLASS/TensorRT GEMM kernels are optimized for those, which leaves a 4-of-16 NoPE tail per head. The measurements below show no accuracy cost from that tail.

## Refactor folded in

`PTv3EncoderExportBase` and `build_monolithic_export_inputs` replace the encoder half that the segmentation, detection and joint export modules each duplicated. That is where most of the deletions come from, and it is why the gating could be added without touching the task models at all.

## Results

Three models trained under an identical recipe at the `ptv3-tiny` scale — same dataset, transforms, optimizer, schedule, `enc_depths` `[2,2,2,2,2]`, `patch_size` 512, 30 epochs each. `h16` keeps ptv3-tiny's head counts (head_dim 16) and is what ships as `litept/tiny`; `h32` halves them (head_dim 32) and is not shipped as a config.

### Accuracy

Evaluated through the **deployed TensorRT engines** via `autoware_ptv3_lib` — not the training checkpoint — over 483 samples / 136.7 M valid points:

| | ptv3-tiny | LitePT h32 | LitePT h16 |
| --- | --- | --- | --- |
| mIoU | **0.7260** | 0.7253 | 0.7253 |
| accuracy | 0.9491 | 0.9483 | 0.9482 |
| params | 31.7 M | **10.5 M** | **10.5 M** |

**LitePT matches PTv3 within 0.07 pp of mIoU at a third of the parameters.** The training-side metrics agree independently (0.6905 / 0.6893 / 0.6897), through different metric code, different preprocessing and PyTorch instead of TensorRT.

### Latency and memory

RTX 4000 SFF Ada, 7 replicates × 50 iterations, replicate means ± sd:

| | ptv3-tiny | LitePT h32 | LitePT h16 |
| --- | --- | --- | --- |
| GPU total | 23.412 ±0.097 ms | 14.460 ±0.057 ms | **13.540 ±0.087 ms** |
| inference | 20.713 ±0.104 ms | 11.909 ±0.037 ms | **10.974 ±0.048 ms** |
| speedup | — | 1.62× | **1.73×** |
| encoder engine | 60.5 MB | **22.1 MB** | 22.1 MB |
| peak GPU | 7841 MiB | **6857 MiB** | 6855 MiB |

Preprocess is unchanged (~2.2 ms) — the node generates the same metadata for all three; the whole gain is inference. Encoder plugin counts drop from `GetIndicePairs ×5 / ImplicitGemm ×10 / Argsort ×2` to `×3 / ×6 / ×0`, with `Cos`/`Sin` ×4 appearing for RoPE.

Caveats: 30 epochs only.


autoware_ml/models/segmentation3d/encoders/ptv3.py
```diff
@@ -308,6 +338,125 @@ def forward(self, coord: torch.Tensor) -> torch.Tensor:
         return output.permute(0, 3, 1, 2)
 
 
+def rotary_span(head_dim: int) -> int:
```
mojomex: Rename to `rope_span`

autoware_ml/models/segmentation3d/encoders/ptv3.py
```diff
@@ -308,6 +338,125 @@ def forward(self, coord: torch.Tensor) -> torch.Tensor:
...
+    """Return how many leading head dimensions an axis-split 3D RoPE can rotate.
+
+    The head dimension is split into three equal per-axis chunks, and each
+    chunk pairs dimension ``i`` with ``i + chunk // 2``. The chunk width must
+    therefore be even, so the rotated span is the largest multiple of six that
+    fits. Note that divisibility by three alone is *not* sufficient: an odd
+    chunk width mispairs the halves and silently stops being a rotation.
```
mojomex: Too much prose, simply mention the three equal per-axism chunks, the i and i + chunk // 2 pairs and that the dim thus has to be a multiple of 6.

autoware_ml/models/segmentation3d/encoders/ptv3.py
```diff
@@ -308,6 +338,125 @@ def forward(self, coord: torch.Tensor) -> torch.Tensor:
...
+    return (head_dim // 6) * 6
+
+
+class Point3DRoPE(nn.Module):
+    """Apply axis-split 3D rotary position embedding to attention queries/keys.
+
+    Positions come from integer voxel coordinates rather than metric ones, so
+    the encoding is exactly reproducible in the exported graph, where per-stage
+    ``grid_coord`` is already part of the input contract while mean-pooled
+    metric coordinates are not.
```
mojomex: Please remove.

autoware_ml/models/segmentation3d/encoders/ptv3.py
```diff
@@ -308,6 +338,125 @@ def forward(self, coord: torch.Tensor) -> torch.Tensor:
...
+
+    Any trailing ``head_dim % 6`` dimensions are passed through unrotated. That
+    keeps tensor-core-friendly head dimensions usable - at ``head_dim=32`` only
+    two of thirty-two channels go unrotated - and partial rotation is standard
+    practice in rotary-embedding implementations.
```
mojomex: Remove "- at ``head_dim ..." and the rest of the paragraph.

autoware_ml/models/segmentation3d/encoders/ptv3.py
```diff
@@ -1001,14 +1221,23 @@ def __init__(
             stem_kernel_size: Embedding-stem submanifold-conv kernel size.
             stem_type: Embedding-stem variant, ``"conv"`` (default) or
                 ``"linear"``. See :class:`Embedding`.
+            enc_conv: Per-stage flag for the submanifold-convolution positional
+                encoding. ``None`` enables it everywhere, which is the PTv3
+                default.
+            enc_attn: Per-stage flag for attention and its MLP. ``None`` enables
+                it everywhere.
```
mojomex: Making these default `True` could be more easily understandable and afaik would lead to the same resulting graph thanks to `expand_stage_flags`.

autoware_ml/models/segmentation3d/encoders/ptv3.py
```diff
@@ -1113,6 +1345,108 @@ def export_forward(self, data_dict: dict[str, torch.Tensor]) -> Point:
...
 
+class LitePTEncoder(PointTransformerV3Encoder):
+    """Configure the PTv3 encoder as LitePT: convolution early, attention late.
+
+    LitePT keeps PTv3's serialization, pooling, and export contract and changes
+    only which operations run at which resolution. Early stages are pure
+    submanifold convolution and late stages are pure attention with rotary
+    position embedding, so the two subsystems never share a block. Because it
+    subclasses the PTv3 encoder, every PTv3 task model, head, and export path
+    accepts it unchanged.
```
mojomex: More concise please "LitePT is a PTv3-compatible model that reduces latency by turning off early attn and late conv blocks. It is fully compatible with PTv3's export contract, encoder, and task heads."

autoware_ml/tests/models/test_litept.py
```diff
@@ -0,0 +1,350 @@
...
+
+@REQUIRES_SPARSE_CUDA
+def test_exported_encoder_graph_matches_the_declared_contract(tmp_path) -> None:
+    """The exported graph must accept exactly the inputs the contract declares.
+
+    A hardcoded expected-name list cannot catch this: it asserts what the contract
+    function returns, not what survives tracing. The exporter prunes inputs the
+    traced model never consumes, so a gated configuration - where a tensor feeds
+    only a disabled branch - silently ends up with an artifact narrower than its
+    own declared interface.
```
mojomex: This is now outdated, please see if this test is still necessary.

autoware_ml/utils/deploy.py
```diff
@@ -385,6 +386,18 @@ def export_to_onnx(
...
+    # artifact narrower than the interface just declared. Restore them so the graph
+    # always accepts exactly `input_names`. Runs on the final file, after any
+    # external-data merge.
+    restored = restore_pruned_graph_inputs(output_path, input_names, input_sample, dynamic_axes)
+    if restored:
+        logger.info(
+            "Restored %d declared input(s) the exporter pruned as unused: %s",
+            len(restored),
+            ", ".join(restored),
+        )
```
mojomex: Also obsolete now

autoware_ml/utils/onnx_modifiers.py
```diff
@@ -251,3 +252,59 @@ def modify(self, onnx_path: str | Path) -> Path:
         ).modify(patched_path)
 
     __call__ = modify
+
+
+def restore_pruned_graph_inputs(
```
mojomex: Obsolete now

docs/models/litept.md
```diff
@@ -0,0 +1,133 @@
...
+icon: lucide/feather
+---
+
+# LitePT
+
+LitePT is a LiDAR encoder that keeps PointTransformerV3's serialization,
+pooling, and export contract and changes only **which operations run at which
+resolution**. Early stages are pure submanifold convolution, late stages are
+pure attention with rotary position embedding, and the decoder is unpooling
+only. The two expensive subsystems never share a block.
```
mojomex: Change prose similarly to what I suggested above.

docs/models/litept.md
```diff
@@ -0,0 +1,133 @@
...
+unrotated. That keeps tensor-core-friendly head dimensions usable: at
+`head_dim = 32`, thirty dimensions rotate and two form a NoPE tail. Partial
+rotation is standard practice in rotary implementations, and the graph cost is
+unchanged either way because the rotation is emitted as one gather plus two
+elementwise operations over the full head dimension, with `cos = 1`/`sin = 0`
+across the tail.
+
+The default `enc_rope_base` of 100.0 mirrors what LitePT's reference CUDA
+operator calls `rope_freq`. The two have **not** been matched numerically, so a
+checkpoint trained against that operator needs verification before reuse.
```
mojomex: Much less prose please.

docs/models/litept.md
mojomex, line 1: Where sensible, refer to ptv3 page for details that are identical across the two models.

docs/models/litept.md
```diff
@@ -0,0 +1,133 @@
...
+bindings from the artifact first.
+
+!!! note "Patch size at the coarsest stages"
+
+    Attention runs only at 1/8 and 1/16 resolution, where the voxel count can
+    approach `enc_patch_size`. `SerializedAttention` refuses to export when a
+    sample has fewer serialized points than the patch size, even though the
+    export-mode padding handles that case. Measure the stage-3 and stage-4
+    voxel counts on the sparsest frames you deploy against and set
+    `enc_patch_size` accordingly.
```
mojomex: Non-issue with the related PR, please remove.

docs/models/litept.md
```diff
@@ -0,0 +1,133 @@
...
+permits and `onnx.checker` accepts. They carry no nodes, cost nothing to
+execute, and the node keeps generating the full per-stage metadata anyway.
+
+The head is the one place where the graph really is smaller: with `dec_depths`
+all zero the segmentation head reduces to per-stage features plus pooling
+clusters, which the existing `dec_depths` rule already covers.
+
+Narrowing the declared contract to what a gated model actually reads is the
+honest end state and remains future work; it needs the runtime to derive its
+bindings from the artifact first.
```
mojomex: Much less prose / refer to PTv3 page.

mojomex:
@amadeuszsz @ktro2828 Thank you for your patience - this is now ready for review.

One design question I'd like to clear: LitePT doesn't have attention at every stage, so some of the pre-computed serialization inputs are not consumed by LitePT. ONNX export automatically prunes them. 

However, the inference node computes all of the inputs, and expects the TRT engine to accept them, so for now, instead of touching the inference node, I added a piece of code that re-adds the unused inputs to satisfy the inference node.

Performance and memory-wise this should be negligible, but I'd like to implement this properly in the inference node at a later time - is it ok to have this workaround in autoware-ml for now?

ktro2828:
@mojomex Thanks for your great work!
I feel it's OK to keep the current inference node for now, and it would be nice if we can design proper implementation later! (for example, defining input adaptor that prepares specific model's intputs)

amadeuszsz:
@codex review

amadeuszsz (changes requested):
One of these:
- Please use modifier (deploy.onnx.modules.<module>.modify_graph argument) instead of using additional script.
- Adjust inference side to handle it.

mojomex:
@amadeuszsz @ktro2828 FYI: I'm doing the proper inference-side implementation now:
* https://github.com/autowarefoundation/autoware_universe/pull/13290

So, the `restore_pruned_graph_inputs` will become obsolete once I open that PR, and I'll remove it - please feel free to skip review of it.

mojomex:
@amadeuszsz The extra machinery restoring the (unused) PTv3 inputs is gone now, please have another look.

autoware_ml/configs/tasks/segmentation3d/litept/seg_head.yaml
```diff
@@ -0,0 +1,25 @@
+# @package model.seg3d_head
+_target_: autoware_ml.models.segmentation3d.heads.ptv3.PTv3SegDecoderHead
+num_classes: ${dataset.segmentation3d.num_classes}
+ignore_index: ${dataset.segmentation3d.ignore_index}
+order: ${model.encoder.order}
+enc_channels: ${model.encoder.enc_channels}
+dec_depths: [0, 0, 0, 0]
```
amadeuszsz: Could you please align with our ptv3's available configurations? For example, there are more parameters which can be inherited instead of rewritten. Also would be nice to have two configurations for `base` and `tiny` versions for now, here (`segmentation3d`) and in `multi`.

mojomex: Addressed in daf51b1 - LitePT now mirrors ptv3's structure with `base` and `tiny`, in `segmentation3d` and in `multi`, and each file carries only what differs from the ptv3 config it composes on top of.

- `base` is LitePT's published parametrization.
- `tiny` is ptv3-tiny with the gating applied and nothing else changed, so the two stay directly comparable.
- The dataset configs (`segmentation3d/litept/voxel012_122m_t4dataset_j6gen2` and the new `multi/litept/...`) are at `tiny` scale, following the ptv3 configs they inherit from.

amadeuszsz (approved):
LGTM

PR #95, opened 2026-08-25
perf(ptv3): fuse mixed-precision TransHead attention

mojomex:
## Motivation

PTv3's TransHead is currently exported as an explicit attention calculation instead of a TensorRT-fusable multi head attention (MHA) pattern.

This is a correctness workaround: TensorRT's fused fp16 attention path produces NaNs, while the workaround which adds `ReduceMax -> Sub` ops before `Softmax` is numerically stable.

However, these added nodes prevent TensorRT from recognizing `MatMul -> Softmax -> MatMul` as MHA, preventing fusion and making the detection head substantially slower and more memory intensive.

The source of the instability is the size of cross-attention: each of the 500 decoder queries attends to 65,536 BEV keys. After the usual maximum subtraction, many exponentials can be close to one, so an online softmax's running normalizer can approach the key count. 65,536 is already larger than fp16's maximum finite value of 65,504, causing the fused fp16 kernel's state to overflow and propagate NaNs.

The bf16 `_gemm_mha_v2` path used here keeps bf16 operands but uses fp32 accumulation for the attention reductions. Together with bf16's 8-bit exponent, which has the same range as fp32, this keeps the softmax state and attention accumulation out of the fp16 overflow regime. Only the Q/K/V attention core is converted: its result is cast back to fp16, so the projections, weights, and rest of the decoder remain fp16.

Self-attention on the other hand has only 500 keys, far below this failure regime, and the fused fp16 self-attention path produced finite outputs in the same validation. We keep it in fp16 to avoid unnecessary lossy casts.

The mixed (BF16 cross, FP16 self attention) path was validated on TensorRT 10.8, 10.10, and 10.16. Production targets need to be SM80 to support bf16 (all of ours are >= SM80).

## Summary

This PR is stacked on [tier4/autoware-ml#89](https://github.com/tier4/autoware-ml/pull/89) and contains only the TransHead performance change.

- Reduce PTv3 inference latency by exporting a TensorRT-fusable TransHead attention graph.
- Add `TransFusionHead.use_bf16_cross_attention` as an explicit export switch. Self-attention stays fp16; the 65,536-key cross-attention core uses bf16 and casts its result back to the surrounding dtype.
- Mark the optimized head as requiring `deploy.onnx.precision=fp16` and refuse export immediately when the model and export settings do not match.
- Keep the existing stabilized, non-fusable attention graph unchanged as the default.
- Add graph regressions for both paths: the optimized path must export direct `MatMul -> Softmax -> MatMul`, while the default must retain `ReduceMax -> Sub -> Softmax -> Cast -> MatMul`.
- Document the paired model/export settings and SM80 requirement.

## Related Issues

- Depends on [tier4/autoware-ml#89](https://github.com/tier4/autoware-ml/pull/89).

## Checklist

## Additional Notes

Enable the optimized path with both settings:

```yaml
model:
  bbox_head:
    use_bf16_cross_attention: true
deploy:
  onnx:
    precision: fp16
```

The bf16 attention core requires SM80 or newer.

## Performance

Nsys shows fused kernels for both self and cross attention now:

image

Measured on ADM-AL30 using a fixed-input PTv3 microbenchmark. Both `segmentation3d.use_head` and `detection3d.use_head` were enabled. The comparison used TensorRT 10.8.0 on an NVIDIA RTX 4000 SFF Ada, with 7 fresh-process replicates, 10 warmup iterations, and 50 measured iterations per replicate (350 measured iterations per engine). Values are the mean of replicate means with a 95% confidence interval.

| GPU metric | Explicit default | Fused fp16/bf16 | Latency reduction |
| --- | ---: | ---: | ---: |
| PTv3 total | 38.995 +/- 0.235 ms | 31.991 +/- 0.279 ms | 18.0% |
| TensorRT inference (enc+seg3d+det3d) | 36.185 +/- 0.126 ms | 28.922 +/- 0.117 ms | 20.1% |
| TensorRT inference (det3d) | 15.168 ms | 7.871 ms | 48.1% |

Steady-state process VRAM was sampled through NVML during three additional runs of 10 warmup and 200 measured iterations. The peak was identical in all three runs:

| GPU memory metric | Explicit default | Fused fp16/bf16 | Reduction |
| --- | ---: | ---: | ---: |
| Full PTv3 process peak VRAM | 10,694 MiB | 9,982 MiB | 712 MiB (6.7%) |
| TensorRT execution context (det3d) | 1,116 MiB | 404 MiB | 712 MiB (63.8%) |

The absolute process footprint includes the CUDA context, all three PTv3 engines, and benchmark buffers. The 712 MiB reduction comes entirely from the detection-head execution context.

The explicit engine contains no MHA layers; the optimized engine contains fused fp16 and bf16 `_gemm_mha_v2` layers.

The isolated TensorRT 10.8 detection-head benchmark (`--useCudaGraph --noDataTransfers`) shows where the full PTv3 gain comes from:

| Detection-head graph | Mean GPU time | Activation memory |
| --- | ---: | ---: |
| Explicit attention | 15.17 ms | 1,169.9 MB |
| Fused fp16/bf16 attention | 7.88 ms | 203.8 MB |

That is a 48.1% latency reduction and an 82.6% engine activation-memory reduction for the measured shape.

### DIFF
--- a/docs/user-guide/deployment.md
+++ b/docs/user-guide/deployment.md
@@ -4,13 +4,18 @@ icon: lucide/package
 
 # Deployment
 
-Autoware-ML exports trained models for production use. The pipeline converts PyTorch checkpoints to optimized inference formats.
+Autoware-ML exports trained models for production use. The pipeline converts
+PyTorch checkpoints to optimized inference formats, one deployable module at a
+time, and stamps every exported module with its identity and provenance.
 
 ## Deployment Pipeline
 
 ```text
-Checkpoint (.ckpt) -> ONNX (.onnx) -> TensorRT (.engine)
+Checkpoint (.ckpt) -> ONNX (.onnx) -> graph modifiers -> precision -> metadata stamp -> TensorRT (.engine)
 ```
+
+Each stage runs per module. The stamp is applied to the final ONNX, so the
+shipped file always carries its own metadata.
 
 ## Basic Usage
 
@@ -25,10 +30,15 @@ parameters to the export model. For a single-task export, pass one
 checkpoint. For a multi-head export, pass one `--weights` per source
 checkpoint (see [Multi-head exports](#multi-head-exports)).
 
-This generates ONNX (`.onnx`) and TensorRT (`.engine`) files when both stages
-are enabled and supported by the model. The deploy command also creates a
-dedicated MLflow run linked to the source training run and logs exported
-artifacts there.
+!!! warning
+    Pass `--release vMAJOR.MINOR.PATCH` if the model is version controlled.
+    Without it the artifacts are stamped `unversioned`, which is fine for
+    quick iteration but never for a release build.
+
+This generates one ONNX (`<module>.onnx`) and one TensorRT engine
+(`<module>.engine`) per deployable module, when both stages are enabled and
+supported by the model. The deploy command also creates a dedicated MLflow run
+linked to the source training run and logs exported artifacts there.
 
 You can disable either stage during iteration:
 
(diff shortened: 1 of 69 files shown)


docs/user-guide/deployment.md
```diff
@@ -157,6 +157,20 @@ Engines are always built strongly typed: the builder uses exactly the precisions
...
+model:
+  bbox_head:
+    use_bf16_cross_attention: true
+deploy:
+  onnx:
+    precision: fp16
+```
+
+Both settings are required, and the resulting model requires an SM80 or newer GPU.
+
```
amadeuszsz: deployment.md documents overall deployment procedures. This section is PTv3-related only, so I don't think we should put it here. Please move it to ptv3.md or remove.

amadeuszsz (changes requested):

docs/user-guide/deployment.md
mojomex: Addressed (moved+reworded) in 5326861

amadeuszsz (approved):
LGTM!

PR #97, opened 2026-08-25
feat(configs): add merge_lists and seg_class_mapping resolvers

amadeuszsz:
## Summary

Two OmegaConf resolvers used by the upcoming metrics and taxonomy configs, plus reworked resolver docs and tests.

- `merge_lists` concatenates interpolated lists, so a config can append task-specific metric components to a shared component list.
- `seg_class_mapping` builds the raw-name to class-index mapping from `name_mapping`, `class_names` and `ignore_index`, failing loud on a final class missing from `class_names`.
- Resolver docstrings rewritten to the documentation policy, tests reworked accordingly.

## Related Issues

Stacked on #96.

## Checklist

- [x] You've checked our [contribution overview](https://tier4.github.io/autoware-ml/contributing/contribution-overview/).
- [x] Your PR follows our [pull request guidelines](https://tier4.github.io/autoware-ml/contributing/pull-request-guidelines/).

## Additional Notes

Pure config-layer groundwork, no runtime behavior changes until later PRs reference the resolvers.

## Additional Log and Media

`pytest autoware_ml/tests/configs autoware_ml/tests/transforms autoware_ml/tests/datamodule autoware_ml/tests/metrics`: 264 passed at this branch head.

PR #98, opened 2026-08-25
feat(metrics): filter and taxonomy axes for the detection framework

amadeuszsz:
## Summary

Reworks the detection metric framework into a suite of injectable components evaluated along filter, range and taxonomy axes.

- Core: `MetricSuite` owns accumulation and DDP sync and dispatches per (filter, range) group; `Metric` components are stateless strategies declaring their stages, filter and required eval keys; behaviour-group taxonomy folding via `class_groups`.
- Matching engine: greedy score-ordered center/corner-distance matching prepared once per class and memoized on a shared `DetectionState`, so every component reads the same match curves.
- Components: mean AP, heading AP, NDS and TP errors on the shared curves, plus six driving-aware additions (corner displacement, heading-flip rate, signed nearest-surface error, calibration error, confident-error rate, matched-only confusion matrix).
- The existing segmentation suite is adapted to the new `state_for` contract, its rework follows later in this stack.
- Detection tests reworked to the new API, new component and grouped-taxonomy tests added.

## Related Issues

Stacked on #97.

## Checklist

## Additional Notes

Behavior notes for reviewers: `MeanAP` reports every configured class at val (NaN-backed, monitored key always present), NDS TP errors use the single `tp_threshold` operating point, test-only components no longer exist at val, and log keys gain filter prefixes and grouped namespaces. Current configs keep composing (the suite-level `thresholds` kwarg moved into the components).

## Additional Log and Media

`pytest autoware_ml/tests/{metrics,configs,datamodule,transforms,preprocessing}`: 285 passed at this branch head. All bundled t4 and nuScenes det/seg/multi metric configs instantiate.

### DIFF
--- a/autoware_ml/metrics/base.py
+++ b/autoware_ml/metrics/base.py
@@ -22,9 +22,22 @@ from dataclasses import dataclass
 from enum import Enum
 from typing import Any, Generic, TypeVar
 
+import numpy as np
+import torch.distributed
 import torchmetrics
 
 logger = logging.getLogger(__name__)
+
+
+def rank_zero() -> bool:
+    """True on the single process that should emit log lines (or when not DDP).
+
+    Returns:
+        Whether this process should log.
+    """
+    if torch.distributed.is_available() and torch.distributed.is_initialized():
+        return torch.distributed.get_rank() == 0
+    return True
 
 
 class EvalStage(str, Enum):
@@ -49,44 +62,162 @@ class MetricRange:
     max_distance: float | None
 
 
+def number_token(value: float) -> str:
+    """Key-safe token for a plain number, e.g. ``99p5`` for 99.5.
+
+    Args:
+        value: Number to encode.
+
+    Returns:
+        The token, with ``.`` written as ``p`` and a ``minus`` prefix for negatives.
+    """
+    return f"{float(value):g}".replace("-", "minus").replace(".", "p")
+
+
 def _distance_token(distance: float) -> str:
-    token = f"{float(distance):g}".replace("-", "minus").replace(".", "p")
-    return f"{token}m"
+    return f"{number_token(distance)}m"
 
 
 def range_suffix(metric_range: MetricRange) -> str:
-    """Collision-free key suffix for a range, e.g. ``0m_50m`` or ``90m_inf``."""
+    """Collision-free key suffix for a range, e.g. ``0m_50m`` or ``90m_inf``.
+
+    Args:
+        metric_range: Radial distance window in meters.
+
+    Returns:
+        The ``<min>m_<max>m`` suffix, an unbounded maximum encoded as ``inf``.
+    """
     lower = _distance_token(metric_range.min_distance)
     if metric_range.max_distance is None:
         return f"{lower}_inf"
     return f"{lower}_{_distance_token(metric_range.max_distance)}"
 
 
+class MetricFilter(ABC):
+    """An optional, per-metric selector applied before a metric runs.
+
+    A filter is an *axis*, not a metric: it decides which accumulated elements
+    (detection boxes, segmentation points) a metric sees, exactly as the range
+    bins do. ``name`` prefixes the metric's keys (empty for the identity filter).
+    ``cache_key`` groups components that share a filter so the suite builds the
+    filtered sub-state once. ``required_eval_keys`` names the per-frame data the
+    filter needs (e.g. the ego pose), folded into the suite's fail-loud check.
+    """
+
+    name: str = ""
+    required_eval_keys: tuple[str, ...] = ()
+
+    @property
+    def cache_key(self) -> str:
+        """Value key grouping equivalent filters. Identity groups share ``""``."""
+        return self.name
+
+    @abstractmethod
+    def keep(self, xyz: Any, context: dict[str, Any]) -> Any:
+        """Boolean mask over elements to retain.
+
+        Args:
+            xyz: Bare base_link points ``(N, 3)`` or full detection box rows
+                ``(N, 7+)``. Concrete filters dispatch on the column count so a
+                box's whole footprint (not just its center) can be tested.
+            context: Per-frame values named in ``required_eval_keys`` (e.g.
+                ``ego2global``, ``scene_token``), supplied by the suite from the
+                model's eval-output.
+
+        Returns:
+            Boolean mask over the input elements, true for those to keep.
+        """
+
+    def available(self, context: dict[str, Any]) -> bool:
+        """Whether this filter can evaluate the given frame.
+
+        A filter that needs external data absent for some frames (e.g. a lanelet
+        map missing for a scene) returns ``False`` so the suite excludes that
+        frame from this filter's bucket (the unfiltered metrics still see it)
+        and counts the exclusion. The default filter always applies. This is an
+        explicit availability check by design: a resource that *should* exist but
+        is corrupt must still fail loud inside ``keep``, never be caught here.
+
+        Args:
+            context: Per-frame values named in ``required_eval_keys``.
+
+        Returns:
+            Whether the frame enters this filter's bucket.
+        """
+        return True
+
+
+class IdentityFilter(MetricFilter):
+    """The default filter. Keeps everything, adds no key prefix."""
+
+    name = ""
+
+    def keep(self, xyz: Any, context: dict[str, Any]) -> Any:
+        """All-true mask over the given elements.
+
+        Args:
+            xyz: Point or box rows ``(N, ...)``.
+            context: Per-frame values, unused here.
+
+        Returns:
+            Boolean mask of ``N`` true values.
+        """
+        return np.ones((len(xyz),), dtype=bool)
+
+
+IDENTITY = IdentityFilter()
+
+
 StateT = TypeVar("StateT")
 
 
 class Metric(ABC, Generic[StateT]):
-    """One injectable metric. Owns its computation; holds no accumulated state.
+    """One injectable metric. Owns its computation and holds no accumulated state.
 
     A metric is a stateless strategy: it reads the synced ``state`` the suite
     builds and returns its slice of the report. ``stages`` declares when it runs
     and is configurable, so the same metric can be light at validation and full
-    at test purely from config.
+    at test purely from config. ``required_eval_keys`` names any extra keys the
+    metric needs in the model's eval-output, folded into the suite's own
+    required-key check so a misconfigured metric fails loud at the first batch.
+    ``filter`` is an optional region/selection axis (default ``None`` means
+    whole-scene) by which the suite groups components and prefixes their keys.
+    ``needs_ttc`` declares that the metric reads the per-box collision TTC, so
+    a suite only runs its collision provider at stages where such a metric is
+    active (TTC is expensive and test-only in practice).
     """
 
-    def __init__(self, stages: tuple[str, ...] | list[str] = ("val", "test")) -> None:
-        """Store the stages this metric runs in.
+    required_eval_keys: tuple[str, ...] = ()
+    needs_ttc: bool = False
+
+    def __init__(
+        self,
+        stages: tuple[str, ...] | list[str] = ("val", "test"),
+        filter: MetricFilter | None = None,
+    ) -> None:
+        """Store the stages this metric runs in and its optional filter.
 
         Args:
             stages: Stage names this metric reports for. Each must name an
-                :class:`EvalStage` value (``"val"`` or ``"test"``); an unknown
+                :class:`EvalStage` value (``"val"`` or ``"test"``), an unknown
                 name raises ``ValueError``.
+            filter: Optional selection axis (e.g. a lanelet ``RegionFilter``).
+                ``None`` means the whole-scene metric.
         """
         self.stages: frozenset[EvalStage] = frozenset(EvalStage(stage) for stage in stages)
+        self.filter: MetricFilter = filter if filter is not None else IDENTITY
 
     @abstractmethod
     def evaluate(self, state: StateT, stage: EvalStage) -> dict[str, float]:
-        """Compute this metric's keys from the suite's synced ``state``."""
+        """Compute this metric's keys from the suite's synced ``state``.
+
+        Args:
+            state: State for one (filter, range) bucket.
+            stage: Evaluation stage being reported.
+
+        Returns:
+            Metric keys mapped to values.
+        """
 
 
 class MetricSuite(torchmetrics.Metric, ABC, Generic[StateT]):
@@ -110,6 +241,7 @@ class MetricSuite(torchmetrics.Metric, ABC, Generic[StateT]):
         self,
         components: list[Metric[StateT]],
         ranges: tuple[MetricRange, ...] = (),
+        prefix: str | None = None,
         **kwargs: Any,
     ) -> None:
         """Store the injected component metrics and ranges.
@@ -119,13 +251,22 @@ class MetricSuite(torchmetrics.Metric, ABC, Generic[StateT]):
                 Empty logs nothing, which is almost always a misconfiguration.
             ranges: Radial windows. Every metric key is also emitted per range
                 with a distance suffix.
+            prefix: Overrides the class-level log prefix. Needed when two suites
+                of the same class run side by side (e.g. an occlusion-split
+                detection suite) so their keys land in distinct namespaces.
+                The eval mixin fails loud on a prefix collision.
             **kwargs: Forwarded to ``torchmetrics.Metric`` (for example
                 ``sync_on_compute``).
         """
         super().__init__(**kwargs)
+        if prefix is not None:
+            self.prefix = str(prefix)
         self.components = list(components)
         self.ranges = tuple(ranges)
         self._stage = EvalStage.TEST
+        # Availability-gated filters (lanelet region / collision area) register here so
+        # the base can log per-filter frame coverage, empty means nothing to gate.
+        self._coverage_filters: list[MetricFilter] = []
         if not self.components:
             logger.warning(
                 "%s was constructed with no components, so it will log nothing.",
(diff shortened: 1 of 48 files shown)


autoware_ml/metrics/detection3d/suite.py
```diff
@@ -109,41 +249,199 @@ def update(self, eval_out: dict[str, Any]) -> None:
...
-            for pred_boxes, pred_scores, pred_labels, gt_boxes, gt_labels in zip(
-                self.pred_boxes,
-                self.pred_scores,
-                self.pred_labels,
-                self.gt_boxes,
-                self.gt_labels,
-                strict=True,
+            if column is not None:
+                gt_mask = self.gt_region_masks[index][:, column].cpu()
+                pred_mask = self.pred_region_masks[index][:, column].cpu()
```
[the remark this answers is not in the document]
amadeuszsz: Fixed. The suite now stores a per-frame availability row next to the keep-masks and `state_for` omits unavailable frames from a filtered slice entirely, so per-frame denominators (e.g. `confident_errors_per_frame`) only count covered frames, matching the coverage contract. Regression assertion added to the map-less-scene suite test.

autoware_ml/metrics/detection3d/calibration.py
```diff
@@ -0,0 +1,92 @@
+"""Detection calibration error (metric E1, detection variant).
```
mojomex: `E1` and other short labels are gone from the reports, please also remove here (multiple sites)

autoware_ml/metrics/detection3d/confident_error.py
```diff
@@ -0,0 +1,66 @@
...
+        # no consumer ever sees. Without a floor the rate tracks the head's score
+        # cutoff more than its confident-error behaviour.
+        self.min_score = float(min_score)
+        if not 0.0 <= self.min_score <= self.score_threshold <= 1.0:
+            raise ValueError("expected 0 <= min_score <= score_threshold <= 1.")
+
+    def evaluate(self, state: DetectionState, stage: EvalStage) -> dict[str, float]:
+        """Report the confident false-positive rate over all classes."""
+        false_positive_total = 0
+        confident_total = 0
```
mojomex: This could be read as "all confident predictions", not just "confident false positives". Please rename to `confident_fals_positives_total` or similar.

autoware_ml/metrics/detection3d/criticality.py
```diff
@@ -0,0 +1,104 @@
...
+    pred_scores: np.ndarray,
+    thresholds: tuple[float, ...],
+) -> dict[float, tuple[np.ndarray, np.ndarray]]:
+    """Score-ordered greedy center-distance matching at several thresholds.
+
+    The distance matrix, the candidate order, and the score order are threshold
+    independent, so they are built once and each threshold runs only its own
+    claim pass. Returns ``{threshold: (is_tp, matched_gt)}`` with the arrays
+    aligned to the predictions in their original order (``matched_gt`` is ``-1``
+    for non-matches).
```
mojomex: Please reformat to a standardized docstring format (such that IDEs can parse the `Returns` correctly).

autoware_ml/metrics/detection3d/matching.py
```diff
@@ -114,87 +143,174 @@ def labels_to_evaluate(
...
+            # The first unclaimed candidate in ascending cost order is the
+            # nearest unclaimed ground truth, past the threshold nothing later
+            # can match either.
+            for candidate in self.candidate_rows[index]:
+                if costs[candidate] > threshold:
+                    break
+                flat = offset + candidate
+                if claimed[flat]:
+                    continue
+                true_positive[index] = 1.0
```
mojomex: Is this just a boolean flag whether the pred at index is a TP? If so, please make the dtype bool.

autoware_ml/metrics/detection3d/matching.py
```diff
@@ -114,87 +143,174 @@ def labels_to_evaluate(
...
+        true_positive = np.zeros(num_pred, dtype=np.float64)
+        matched_flat_gt = np.full(num_pred, -1, dtype=np.int64)
+        claimed = np.zeros(self.gt_flat.shape[0], dtype=bool)
+        for index in range(num_pred):
+            costs = self.cost_rows[index]
+            offset = int(self.gt_flat_index[index])
+            # The first unclaimed candidate in ascending cost order is the
+            # nearest unclaimed ground truth, past the threshold nothing later
+            # can match either.
+            for candidate in self.candidate_rows[index]:
```
mojomex: Quickly reading through the code, it's hard to understand what a `candidate` is - please rename or state in a comment. Maybe `candidate_gt`?

autoware_ml/metrics/detection3d/matching.py
```diff
@@ -114,87 +143,174 @@ def labels_to_evaluate(
...
+        nearest_surface = np.full(num_pred, np.nan, dtype=np.float64)
+        matched = matched_flat_gt >= 0
+        if matched.any():
+            pred_boxes = self.pred_boxes[matched]
+            gt_boxes = self.gt_flat[matched_flat_gt[matched]]
+            errors["ATE"][matched] = np.linalg.norm(pred_boxes[:, :2] - gt_boxes[:, :2], axis=1)
+            errors["AOE"][matched] = _orientation_errors(pred_boxes, gt_boxes)
+            errors["ASE"][matched] = _scale_errors(pred_boxes, gt_boxes)
+            errors["AVE"][matched] = _velocity_errors(pred_boxes, gt_boxes)
+            errors["AAE"][matched] = 1.0
```
mojomex, line 332: This is a dummy value right? Please add a `//TODO`.

autoware_ml/metrics/detection3d/matching.py
```diff
@@ -369,12 +486,83 @@ def _mean_or_one(values: np.ndarray) -> float:
...
 
-# Public alias: metrics compute class-means and threshold-means with this.
-mean_valid = _mean_valid
+@dataclass
+class DetectionState:
+    """Synced detection state the suite hands to each metric.
+
+    Attributes:
+        samples: Per-frame samples (already GT-filtered, optionally range-clipped).
+        class_names: Ordered class names for metric keys, or ``None``.
```
mojomex: What happens when set to ``None``?

autoware_ml/metrics/detection3d/matching.py
```diff
@@ -369,12 +486,83 @@ def _mean_or_one(values: np.ndarray) -> float:
...
+            self._metrics_cache[key] = metrics
+        return metrics
+
+    def match_curve(self, label: int, threshold: float) -> MatchCurve:
+        """Return the score-ordered match curve for a class and threshold.
+
+        Memoized, so the matching runs once per ``(label, threshold)`` and is
+        shared by every metric that asks for it.
+        """
+        key = (label, float(threshold))
```
mojomex, line 687: [just fyi] Float keys might be brittle (changes by 0.0...01 -> key changes) - all usages I've seen in this codebase are raw float literals, so identical, but any computed threshold would fall apart.

autoware_ml/metrics/detection3d/mean_ap.py
```diff
@@ -24,15 +43,15 @@ def evaluate(self, state: DetectionState, stage: EvalStage) -> dict[str, float]:
         Returns:
             Mapping of metric names to scalar values.
         """
-        full = stage is EvalStage.TEST
-        labels = state.labels(full)
+        # One class-set rule for both stages, so the model-selection number and
+        # the reported test number average the same classes, and the monitored
+        # key is always present (NaN when nothing is measurable, never absent).
```
mojomex: This comment is kind of hard to parse, please rephrase.

autoware_ml/metrics/detection3d/naming.py
```diff
@@ -5,20 +5,36 @@
...
+    if class_names is None:
+        return f"class_{label}"
+    if not 0 <= label < len(class_names):
+        raise ValueError(f"label {label} is outside the configured {len(class_names)} class names.")
+    return metric_token(class_names[label])
+
+
+def number_token(value: float) -> str:
+    """Key-safe token for a plain number, e.g. ``99p5`` for 99.5."""
+    return f"{float(value):g}".replace("-", "minus").replace(".", "p")
```
mojomex: Seems like it could be collapsed together with `_distance_token`

autoware_ml/metrics/detection3d/heading_ap.py
```diff
@@ -3,19 +3,34 @@
...
 from autoware_ml.metrics.detection3d.naming import label_metric_name, threshold_token
-from autoware_ml.metrics.detection3d.structures import DetectionState
+from autoware_ml.metrics.detection3d.structures import DEFAULT_MATCH_THRESHOLDS
 
 
 class HeadingAP(Metric[DetectionState]):
     """Class-mean APH plus per-class and per-threshold APH. APH weights each true
     positive by its heading score, so it needs the orientation errors the match
-    curve carries.
+    curve carries. Averaged over ``thresholds``.
```
mojomex, line 14: Is there a use case for averaging over multiple thresholds? Is this common practice?

autoware_ml/metrics/detection3d/structures.py
```diff
@@ -1,35 +1,60 @@
...
 
 ERROR_NAMES = ("ATE", "AOE", "ASE", "AVE", "AAE")
 
+# nuScenes center-distance AP thresholds (meters). The AP-family metrics own this
+# set. Metrics that need a single operating point share DEFAULT_TP_THRESHOLD (the
+# nuScenes dist_th_tp), so the whole report moves together when it changes. Nds
+# and TpErrors additionally require tp_threshold to be a configured threshold, so
+# the aggregate operating point always appears among the detail keys.
+DEFAULT_MATCH_THRESHOLDS = (0.5, 1.0, 2.0, 4.0)
+DEFAULT_TP_THRESHOLD = 2.0
```
mojomex, line 24: [future work] For large objects 2m might be sensible, but for small objects of size << 2m (like debris, cones, pedestrians), this becomes meaningless (we match small objects completely disjunct with a GT box to that box anyway).

autoware_ml/tests/metrics/test_metric_components.py
```diff
@@ -87,6 +90,60 @@ def test_tp_error_means_ignore_classes_without_true_positives() -> None:
...
+        gt_boxes=gt,
+        gt_labels=torch.tensor([0]),
+    )
+    return DetectionState(samples=[sample], class_names=("car",))
+
+
+def test_det_confident_error_flags_high_score_false_positive() -> None:
+    out = ConfidentErrorRate(score_threshold=0.5).evaluate(_det_state_tp_fp(), EvalStage.TEST)
+    assert out["confident_error_rate"] == pytest.approx(1.0)  # the one FP (0.8) is confident
+    assert out["confident_error_count"] == pytest.approx(1.0)
```
mojomex, line 116: Counts should be `uint64` (or at least an integer type)

autoware_ml/tests/metrics/test_driving_aware_components.py
```diff
@@ -0,0 +1,97 @@
...
+def test_heading_flip_rate_no_flip_for_small_error() -> None:
+    out = HeadingFlipRate().evaluate(_yaw_state(0.1), EvalStage.TEST)
+    assert out["flip_count_car"] == 0.0
+    assert out["flip_rate_car"] == pytest.approx(0.0)
+
+
+def test_heading_flip_rate_counts_a_reversal() -> None:
+    # A near-180 deg error is a flip: uniform +1 regardless of exact angle.
+    out = HeadingFlipRate().evaluate(_yaw_state(pi - 0.1), EvalStage.TEST)
+    assert out["flip_count_car"] == 1.0
```
mojomex, line 77: Counts should be integers

mojomex (commented):
Partial review (I'm done with 18 of 28 files, so it will take a bit longer 🙏).

autoware_ml/metrics/detection3d/calibration.py
amadeuszsz: Removed here and in the later PRs of the stack. Addressed in https://github.com/tier4/autoware-ml/pull/98/commits/56e99242b5b7ad33e1c944e71b7693646c1b0484.

autoware_ml/metrics/detection3d/confident_error.py
amadeuszsz: Renamed. Addressed in https://github.com/tier4/autoware-ml/pull/98/commits/298fcce8610c7ba76abee8e3233d9bf9df34478e.

autoware_ml/metrics/detection3d/criticality.py
amadeuszsz: Reformatted with Args and Returns sections. Addressed in https://github.com/tier4/autoware-ml/pull/98/commits/9fb6cf6564f86b99e23daf1b4c74e307608f54ca.

autoware_ml/metrics/detection3d/matching.py
amadeuszsz: Right, it is a flag. Both flags are bool now and the consumers use them as plain masks. Addressed in https://github.com/tier4/autoware-ml/pull/98/commits/4292b1eead4682efc299cf1be752a5a8f812f48f.

autoware_ml/metrics/detection3d/matching.py
amadeuszsz: Renamed to candidate_gt. Addressed in https://github.com/tier4/autoware-ml/pull/98/commits/f297a03e13f480bfd450b34ab8c7790ec5eb9ea2.

autoware_ml/metrics/detection3d/matching.py
amadeuszsz: Added. Addressed in https://github.com/tier4/autoware-ml/pull/98/commits/1bde1ead063f9ea1f2265dcc2098f302cada5f9b.

autoware_ml/metrics/detection3d/matching.py
amadeuszsz: Metric keys then fall back to class_<label> tokens. Documented in the docstring. Addressed in https://github.com/tier4/autoware-ml/pull/98/commits/edea8e7f5e0b7e62404cf64ebaf49f53a7267dbc.

autoware_ml/metrics/detection3d/matching.py
amadeuszsz: Right. Thresholds only come from config literals today and float() normalizes int inputs, so the keys stay identical. If a computed threshold ever shows up we should normalize the key precision.

autoware_ml/metrics/detection3d/mean_ap.py
amadeuszsz: Rephrased. Addressed in https://github.com/tier4/autoware-ml/pull/98/commits/e9e457e6b1b84d34996c9474bc3a79f7c23807d3.

autoware_ml/metrics/detection3d/naming.py
amadeuszsz: Done, number_token now lives next to _distance_token in base and the distance rule reuses it. Addressed in https://github.com/tier4/autoware-ml/pull/98/commits/2db487e539db1fd25513dd1b88cf9a5f1c3a28e5.

autoware_ml/metrics/detection3d/heading_ap.py
amadeuszsz: It mirrors our mAP convention, nuScenes averages AP over the 0.5/1/2/4 m matches, so mAPH stays directly comparable to mAP and reuses the same match curves.

autoware_ml/metrics/detection3d/structures.py
amadeuszsz: Agreed. 2 m is the nuScenes dist_th_tp convention kept for comparability, and the AP family averages 0.5 to 4 m which helps a bit, but a size relative threshold is the right follow up.

autoware_ml/tests/metrics/test_metric_components.py
amadeuszsz: The report contract is dict[str, float] end to end, one scalar type through torchmetrics sync and Lightning logging, so counts are whole valued floats and the test asserts through that contract.

autoware_ml/tests/metrics/test_driving_aware_components.py
amadeuszsz: Same as the other count comment, counts ride the float report contract so they stay whole valued floats.

autoware_ml/metrics/detection3d/tp_errors.py
```diff
@@ -26,14 +27,45 @@ class TpErrors(Metric[DetectionState]):
...
+        self.thresholds = tuple(float(threshold) for threshold in thresholds)
+        if not self.thresholds:
+            raise ValueError("thresholds must not be empty.")
+        # Aggregate means use the single nuScenes operating point. The per-class
+        # per-threshold detail keys still cover the full threshold set.
+        self.tp_threshold = float(tp_threshold)
+        if self.tp_threshold not in self.thresholds:
+            raise ValueError(
+                f"tp_threshold={self.tp_threshold} is not one of thresholds={self.thresholds}, "
+                "the aggregate operating point must be an explicitly configured threshold."
```
mojomex, line 57: [future work] same float `==` comment as above

autoware_ml/metrics/detection3d/suite.py
```diff
@@ -45,13 +72,51 @@ def __init__(
...
         self.class_names = tuple(class_names) if class_names is not None else None
-        self.thresholds = tuple(float(threshold) for threshold in thresholds)
+        _validate_name_tokens(self.class_names)
         self.eval_class_range = eval_class_range
         self.min_num_points = int(min_num_points)
+        self.match_cost = str(match_cost)
+        resolve_match_cost(self.match_cost)  # a config typo must fail at build, not after the epoch
+        # Optional collision provider (CollisionTTC). When set, the
+        # per-box time-to-collision to ego is computed once per frame at update and
+        # carried on each sample for the criticality metrics. Requires ego2global + scene_token in the eval output.
```
mojomex: Partially duplicates the docstring above, please de-duplicate

mojomex (dismissed):
**I haven't reviewed everything.**

I've reviewed as far as possible but the PR introduces too many changes at once, to infra I'm already unfamiliar with. Quality of what I've reviewed (about 2/3 of the changed files) seems high and this is not immediately mission-critical code, so I've approved this time.

For the future, let's please start with a design doc, followed by small incremental PRs (even if not polished) so changes don't blow through my head's context window 🙏

autoware_ml/metrics/detection3d/suite.py
amadeuszsz: https://github.com/tier4/autoware-ml/pull/98/commits/f34b397bce4ec2c34e0b401c18d444022bf9b2c9

autoware_ml/metrics/detection3d/suite.py
```diff
@@ -45,13 +72,48 @@ def __init__(
         ),
         eval_class_range: dict[str, float] | None = None,
         min_num_points: int = 0,
+        match_cost: str = "center",
+        class_groups: dict[str, list[str]] | None = None,
+        collision: Any = None,
```
mojomex: `collision: Any` is a duck-typed interface (`available(scene_token)`, `per_box_ttc(boxes, labels, ego2global, scene_token)`) with no definition anywhere in the repo, and no component in this PR reads the TTC it produces. Roughly 120 lines of this file exist to serve it.

Suggestion: defer the TTC states, `_frame_ttc`, `_ttc_frame_covered`, `_log_ttc_coverage` and the TTC fields on `Detection3DSample` to the PR that adds the provider and the criticality metrics. If you'd rather keep it here, please add a `typing.Protocol` so the two method signatures are written down once and a reader can find them.

autoware_ml/metrics/detection3d/suite.py
```diff
@@ -109,41 +257,210 @@ def update(self, eval_out: dict[str, Any]) -> None:
...
+        criticality metrics are test-only must not pay for it every epoch.
+
+        Args:
+            stage: Evaluation stage this instance accumulates and reports for.
+        """
+        super().bind_stage(stage)
+        if self.collision is not None and not any(
+            component.needs_ttc for component in self.active_components()
+        ):
+            self.collision = None
```
mojomex: Since no `Metric` subclass sets `needs_ttc = True` in this commit, this branch always fires and the provider is always dropped, at test too. Confirms the TTC path is unreachable right now. If it stays, a test that a `needs_ttc` component keeps the provider at test and loses it at val would pin the intended behaviour.

autoware_ml/metrics/base.py
```diff
@@ -49,44 +62,162 @@ class MetricRange:
...
+    whole-scene) by which the suite groups components and prefixes their keys.
+    ``needs_ttc`` declares that the metric reads the per-box collision TTC, so
+    a suite only runs its collision provider at stages where such a metric is
+    active (TTC is expensive and test-only in practice).
     """
 
-    def __init__(self, stages: tuple[str, ...] | list[str] = ("val", "test")) -> None:
-        """Store the stages this metric runs in.
+    required_eval_keys: tuple[str, ...] = ()
+    needs_ttc: bool = False
```
mojomex: Same theme: a base-class flag that nothing sets. If the TTC plumbing moves out, this goes with it.

autoware_ml/metrics/base.py
```diff
@@ -137,41 +278,264 @@ def __init__(
...
+        covered = self.region_frames_covered.tolist()
+        for metric_filter, n_seen, n_covered in zip(
+            self._coverage_filters, seen, covered, strict=True
+        ):
+            if not n_seen:
+                continue
+            if n_covered < n_seen:
+                logger.warning(
+                    "Filter %r ran on %d/%d frames, %d frame(s) in scenes without a "
+                    "lanelet map were excluded from its slice (whole-scene metrics keep them).",
```
mojomex: nit: the generic `MetricSuite` shouldn't know about lanelet maps. `MetricFilter.available` is documented as a general availability hook, so this message will be wrong for any other kind of filter. "frames the filter was unavailable for" keeps it honest.

autoware_ml/metrics/detection3d/criticality.py
```diff
@@ -0,0 +1,131 @@
+"""Shared frame-local matching and weighted-AP helpers for the detection metrics.
+
+The reachability time-to-collision and its risk weight live in
+:mod:`autoware_ml.metrics.geometry.reachability`, and the criticality metrics read a per-box TTC the suite has already computed. This module holds only the
```
mojomex: `autoware_ml.metrics.geometry.reachability` doesn't exist in the tree, and there are no criticality metrics in this commit. Meanwhile the only thing imported from this file is `greedy_match` (by `ConfusionMatrix`), which is a second, independent implementation of the greedy matcher next to `PreparedLabel.match`.

Suggestion: move `greedy_match` into `matching.py` next to its sibling, and drop `greedy_match_thresholds` and `weighted_average_precision` (both unused) until the PR that needs them. Then this file disappears and there's one place to look for matching.

autoware_ml/metrics/detection3d/confusion_matrix.py
```diff
@@ -0,0 +1,90 @@
...
+        Args:
+            state: Detection state for one (filter, range) bucket.
+            stage: Evaluation stage being reported.
+
+        Returns:
+            Metric keys mapped to values.
+        """
+        if state.class_names is None:
+            raise ValueError("ConfusionMatrix needs class_names to label its cells.")
+        if state.match_cost != "center":
```
mojomex: This raises at `compute`, i.e. after the whole test epoch has run. The suite's own comment at `resolve_match_cost` says "a config typo must fail at build, not after the epoch", and the suite knows both `match_cost` and its components in `__init__`, so this check (and the `class_names is None` one above) can live there. Same applies to the score-range check in `calibration.py`: a model emitting logits would burn a full test run before failing. Could be checked on the first `update` instead.

autoware_ml/metrics/detection3d/confident_error.py
```diff
@@ -0,0 +1,80 @@
+"""Detection confident-error rate.
+
+Of the boxes the model got wrong, what fraction was it sure about? A high-score
+false positive is the confident phantom that triggers a hard brake, so it is the
```
mojomex: Safety framing check: matching is per class, so a "false positive" here also includes (a) a well-localized box with the wrong class and (b) a duplicate box on an already-claimed GT. Neither is a phantom, and (a) in particular wouldn't cause a spurious hard brake. Either soften the docstring to "confident false positive" without the phantom story, or count FPs class-agnostically (you already have `greedy_match` for that). As-is, a model with class confusion between truck/bus looks like it hallucinates objects.

autoware_ml/metrics/detection3d/nds.py
```diff
@@ -44,24 +74,29 @@ def evaluate(self, state: DetectionState, stage: EvalStage) -> dict[str, float]:
...
+            if selected.count > 0:
+                error_dicts.append(selected.errors)
 
         mean_ap = mean_valid(per_class_ap)
         mean_aph = mean_valid(per_class_aph)
-        errors = mean_tp_errors(error_dicts) if error_dicts else {name: 1.0 for name in ERROR_NAMES}
+        errors = (
+            mean_tp_errors(error_dicts)
+            if error_dicts
+            else {name: float("nan") for name in ERROR_NAMES}
```
mojomex: With NaN errors, `nds()` does `max(0.0, 1.0 - nan)` which returns `0.0` in Python, so `map_based_nds` silently becomes `0.5 * mAP`, a plausible-looking number, while `mATE` etc. correctly read NaN in the same report. Numerically it's what the old `1.0` fallback produced, but now the intent is clearly "not measurable". Suggest returning NaN from `nds()` when any error is NaN so the two agree. Same fallback exists in `tp_errors.py` but there it only feeds the `m*` keys, so it's fine.

autoware_ml/metrics/detection3d/nearest_surface_error.py
```diff
@@ -0,0 +1,86 @@
+"""Signed nearest-surface error.
+
+Stopping distance is computed against the nearest face of an object, not its
```
mojomex: Small caveat worth stating in the docstring: the distance is from the base_link origin (rear axle for most platforms) to the nearest face in any direction, and it's computed for every matched object. The "brakes late / over-caution" reading only holds for objects ahead of ego; for a car alongside or behind, a positive error is just a radial near-face error. Fine as a metric, but the sign semantics shouldn't be oversold.

autoware_ml/metrics/detection3d/geometry.py
```diff
@@ -0,0 +1,217 @@
...
+
+    Args:
+        boxes: Box rows ``(N, 7+)``.
+
+    Returns:
+        Corner coordinates ``(N, 4, 2)``.
+    """
+    boxes = np.asarray(boxes, dtype=np.float64)
+    half_dx, half_dy = boxes[:, 3] / 2.0, boxes[:, 4] / 2.0
+    # Same counter-clockwise corner order as bev_corners.
```
mojomex: nit: `bev_corners` says clockwise, this says counter-clockwise. The order (+,+),(+,-),(-,-),(-,+) is clockwise in an x-right/y-up frame, so this comment is the wrong one.

autoware_ml/tests/metrics/test_detection3d.py
```diff
@@ -404,3 +419,75 @@ def test_detailed_center_distance_metrics_report_optimal_error_subset() -> None:
         torch.tensor(metrics["tp_error_num_match_car_optimal_0p5m"]), torch.tensor(2.0)
     )
     assert torch.isclose(torch.tensor(metrics["optimal_conf_car_0p5m"]), torch.tensor(0.7))
+
+
+def test_match_by_cost_equals_reference_global_greedy() -> None:
```
mojomex: This test is great and is the reason I'm comfortable with the matcher rewrite.

What I couldn't find: any test for `class_groups` (the description says grouped-taxonomy tests were added) or for the filter axis. Minimal set I'd want:
- `resolve_class_groups`: rejects overlap, rejects uncovered class, rejects group name colliding with a trained class, skips an untrained member.
- a grouped suite where a bus predicted as truck is a TP for the group.
- a fake `MetricFilter` with `available()` false on some frames: those frames are absent from the filtered slice but present in the identity slice, keys carry the `name/` prefix, and the coverage counters match.

PR #99, opened 2026-08-25
feat(metrics): lanelet map regions and speed limits

amadeuszsz:
## Summary

Lanelet2 map parsing for the upcoming region-based evaluation filters.

- Parses a scene's `map/lanelet2_map.osm` (standard-library XML + shapely, no lanelet2 dependency) into region polygons keyed by lanelet `subtype` or area way `type`.
- `LaneletMap` answers point-in-region and footprint-overlap membership with prepared geometry, a border margin eroding only the outer border of the mapped surface, and per-lane `speed_limit` lookup.
- `LaneletMapProvider` caches one parsed map per scene with an explicit availability check for map-less scenes; the T4 resolver derives the OSM path from the scene directory fragment.

## Related Issues

Stacked on #98.

## Checklist

## Additional Notes

Purely additive module, nothing references it yet. Its consumer tests arrive with the filters PR next in the stack.

## Additional Log and Media

`pytest autoware_ml/tests/{metrics,configs}`: 72 passed at this branch head.

### DIFF
--- a/autoware_ml/metrics/detection3d/suite.py
+++ b/autoware_ml/metrics/detection3d/suite.py
@@ -3,7 +3,10 @@
 ``Detection3DMetricSuite`` owns the per-frame prediction and ground-truth tensors
 as list states (so torchmetrics handles cross-GPU sync) and applies the GT
 filters at ``update`` time. It knows nothing about which metrics run: it builds
-a ``DetectionState`` (overall and per range) and hands it to the injected metrics.
+a ``DetectionState`` (overall and per range/filter) and hands it to the injected
+metrics. Region and collision filters need per-frame metadata (ego pose, scene token) in the
+eval output, so their keep-masks are precomputed per frame at update and nothing needs the map
+at compute time, DDP gathers only tensors.
 """
 
 from __future__ import annotations
@@ -11,16 +14,41 @@ from __future__ import annotations
 import logging
 from typing import Any
 
+import numpy as np
 import torch
 
-from autoware_ml.metrics.base import Metric, MetricRange, MetricSuite
+from autoware_ml.metrics.base import (
+    IDENTITY,
+    Metric,
+    MetricFilter,
+    MetricRange,
+    MetricSuite,
+    rank_zero,
+)
+from autoware_ml.metrics.class_groups import fold_labels, resolve_class_groups
 from autoware_ml.metrics.detection3d.matching import (
+    DetectionState,
     clip_to_range,
     gt_keep_mask,
+    resolve_match_cost,
 )
-from autoware_ml.metrics.detection3d.structures import Detection3DSample, DetectionState
+from autoware_ml.metrics.detection3d.naming import metric_token
+from autoware_ml.metrics.detection3d.structures import Detection3DSample
 
 logger = logging.getLogger(__name__)
+
+
+def _validate_name_tokens(names: tuple[str, ...] | None) -> None:
+    """Reject class-name sets whose metric-key tokens collide."""
+    if not names:
+        return
+    tokens = [metric_token(name) for name in names]
+    duplicates = sorted({token for token in tokens if tokens.count(token) > 1})
+    if duplicates:
+        raise ValueError(
+            f"class names collapse to duplicate metric-key tokens {duplicates}, "
+            "per-class values would silently overwrite each other."
+        )
 
 
 class Detection3DMetricSuite(MetricSuite[DetectionState]):
@@ -36,7 +64,6 @@ class Detection3DMetricSuite(MetricSuite[DetectionState]):
         self,
         components: list[Metric[DetectionState]],
         class_names: tuple[str, ...] | None = None,
-        thresholds: tuple[float, ...] = (0.5, 1.0, 2.0, 4.0),
         ranges: tuple[MetricRange, ...] = (
             MetricRange("0-50m", 0.0, 50.0),
             MetricRange("50-90m", 50.0, 90.0),
@@ -45,13 +72,48 @@ class Detection3DMetricSuite(MetricSuite[DetectionState]):
         ),
         eval_class_range: dict[str, float] | None = None,
         min_num_points: int = 0,
+        match_cost: str = "center",
+        class_groups: dict[str, list[str]] | None = None,
+        collision: Any = None,
         **kwargs: Any,
     ) -> None:
+        """Validate the class setup and register the accumulation states.
+
+        Args:
+            components: Injected metrics, as in :class:`MetricSuite`.
+            class_names: Ordered class names for metric keys, or ``None`` to report only
+                classes present in the ground truth.
+            ranges: Radial windows every key is also emitted for.
+            eval_class_range: Class name to maximum GT distance in meters, applied as a GT
+                filter at accumulation.
+            min_num_points: Minimum LiDAR points a GT box needs to count, 0 disables the
+                filter.
+            match_cost: Matching cost name, ``center`` or ``corner``.
+            class_groups: Group name to member class names. When set, labels are folded onto
+                the group taxonomy at state build, so matching happens between grouped classes.
+            collision: Optional collision TTC provider. When set, the per-box
+                time-to-collision to ego is computed once per frame at update and carried on
+                each sample for the criticality metrics.
+            **kwargs: Forwarded to :class:`MetricSuite`.
+        """
         super().__init__(components=components, ranges=ranges, **kwargs)
         self.class_names = tuple(class_names) if class_names is not None else None
-        self.thresholds = tuple(float(threshold) for threshold in thresholds)
+        _validate_name_tokens(self.class_names)
         self.eval_class_range = eval_class_range
         self.min_num_points = int(min_num_points)
+        self.match_cost = str(match_cost)
+        resolve_match_cost(self.match_cost)  # a config typo must fail at build, not after the epoch
+        self.collision = collision
+        # When class_groups is set, GT/pred labels are folded onto the behaviour
+        # taxonomy at state build (after the per-class GT filters), so matching
+        # happens between grouped classes: a bus predicted as a truck is a true
+        # positive for grouped_vehicle. The synced boxes stay the trained classes.
+        self._group_lut, self._grouped_names = (
+            resolve_class_groups(self.class_names, class_groups) if class_groups else (None, None)
+        )
+        # The grouped names are what label_metric_name tokenizes in a grouped
+        # suite, so they need the same duplicate-token guard.
+        _validate_name_tokens(self._grouped_names)
 
         if eval_class_range and not self.class_names:
             raise ValueError("class_names must be provided when eval_class_range is configured.")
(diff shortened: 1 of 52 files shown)


autoware_ml/metrics/detection3d/suite.py
```diff
@@ -89,6 +225,11 @@ def update(self, eval_out: dict[str, Any]) -> None:
             raise ValueError(
                 "Detection metric expects equal numbers of predictions, gt_boxes, and gt_labels."
             )
+        if self.min_num_points > 0 and gt_num_points is None:
```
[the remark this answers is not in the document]
amadeuszsz: Good catch, the multi-task collation was indeed missing the point counts. Fixed in #98 (the commit that makes the missing value fatal): `gt_num_points: list` added to the collation map in `configs/tasks/multi/ptv3/base.yaml`.

autoware_ml/metrics/geometry/lanelet.py
```diff
@@ -0,0 +1,490 @@
...
+            return hit
+        if expand:
+            in_expanded = np.asarray(
+                shapely.intersects(self._expanded_region(tuple(tokens), float(margin)), footprints),
+                dtype=bool,
+            )
+            on_map = np.asarray(shapely.intersects(self._full_surface(), footprints), dtype=bool)
+            return hit | (in_expanded & ~on_map)
+        eroded = self._eroded_full_surface(float(margin))
+        return hit & np.asarray(shapely.intersects(eroded, footprints), dtype=bool)
```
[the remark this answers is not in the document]
amadeuszsz: Good catch, the two overlap tests can indeed be satisfied by different parts of one footprint. The eroded slice now tests against a single geometry, the selected region intersected with the eroded surface, cached like the other prepared unions. A regression test with exactly this band straddler rides in #101 where the filter tests live. Addressed in https://github.com/tier4/autoware-ml/pull/99/commits/ff6f91a32e8e277048a371cb0d70052c1fdda365.

autoware_ml/metrics/geometry/lanelet.py
```diff
@@ -0,0 +1,505 @@
+"""Lanelet2 map parsing into region polygons, for the region-filter evaluation axis.
+
+A T4 scene ships ``map/lanelet2_map.osm`` in the same local map frame the ego
+poses use. This module parses it into shapely polygons grouped by lanelet2 token,
+either a lanelet's ``subtype`` (road, walkway, crosswalk, road_shoulder) or an
+area way's ``type`` (drivable_area, crosswalk_polygon, intersection_area). A
+:class:`LaneletMap` then answers point-in-region membership. The map for a scene
+is parsed once and cached.
+
+lanelet2 itself is not installed in the container, so the OSM is parsed with the
```
mojomex: Is there a strong reason for this? It seems `lanelet2` is pip-installable, and it would prevent hard-to-catch format mismatches or edge cases, so would be preferable.

autoware_ml/metrics/geometry/lanelet.py
```diff
@@ -0,0 +1,505 @@
...
+
+    return dict(regions)
+
+
+def _safe_polygon(ring: list[tuple[float, float]]) -> Polygon | None:
+    if len(ring) < 3:
+        return None
+    polygon = Polygon(ring)
+    if not polygon.is_valid:
+        polygon = polygon.buffer(0)  # repair self-touching rings
```
mojomex: Does this fix always make the polygon valid?

autoware_ml/metrics/geometry/lanelet.py
```diff
@@ -0,0 +1,505 @@
...
+    nodes: dict[str, tuple[float, float]],
+    ways: dict[str, ET.Element],
+) -> list[tuple[Polygon, float]]:
+    speeds: list[tuple[Polygon, float]] = []
+    for relation in root.findall("relation"):
+        tags = _tags(relation)
+        if tags.get("type") != "lanelet" or "speed_limit" not in tags:
+            continue
+        try:
+            speed_mps = float(tags["speed_limit"]) / 3.6  # km/h -> m/s
```
mojomex: Failing to convert an existing speed_limit field should fail loudly instead of silently being ignored.

autoware_ml/metrics/geometry/lanelet.py
```diff
@@ -0,0 +1,505 @@
...
+            The speed limit in m/s.
+        """
+        if self._speed_index is None:
+            return default
+        point = Point(x, y)
+        candidates = self._speed_index.query(point)
+        speeds = [
+            float(self._speed_values[i]) for i in candidates if self._speed_polys[i].contains(point)
+        ]
+        return min(speeds) if speeds else default
```
mojomex: * Taking the min here might result in overly optimistic TTC, so maybe taking the `max` or having a param to toggle would be best.
* For complex roads like in Tokyo, there might be lanelets overlapping in (x,y) space because they are on different z levels. Niot a problem now but let's keep it in mind for the future.

autoware_ml/metrics/geometry/lanelet.py
```diff
@@ -0,0 +1,505 @@
...
+        Used by the reachability collision engine to clip wheeled reachable sets
+        to the drivable area (road / road_shoulder / crosswalk).
+
+        Args:
+            tokens: Region tokens to unite.
+
+        Returns:
+            The shapely (multi)polygon union.
+        """
+        return self._region_union(tuple(tokens))
```
mojomex: This function _only_ calls the `_region_union` with identical params, so the two should be collapsed into one.

autoware_ml/metrics/geometry/lanelet.py
```diff
@@ -0,0 +1,505 @@
...
+
+        Returns:
+            Boolean mask ``(N,)``.
+        """
+        if xy.shape[0] == 0:
+            return np.zeros((0,), dtype=bool)
+        points = shapely.points(xy[:, 0], xy[:, 1])
+        mask = np.asarray(shapely.contains(self._region_union(tuple(tokens)), points), dtype=bool)
+        if margin <= 0.0:
+            return mask
```
mojomex: As expand/erode is explicitly handled by a separate bool, negative margins should not be allowed at all.

Might be cleaner to collapse margin/expand into one.

autoware_ml/metrics/geometry/lanelet.py
```diff
@@ -0,0 +1,505 @@
...
+
+        Args:
+            scene_token: Scene identifier.
+
+        Returns:
+            The scene's parsed map.
+        """
+        if scene_token not in self._cache:
+            self._cache[scene_token] = LaneletMap.from_osm(self._resolve_osm(scene_token))
+        return self._cache[scene_token]
```
mojomex: Could this manual caching logic be replaced by a `@lru_cache` decorator?

autoware_ml/metrics/geometry/lanelet.py
```diff
@@ -0,0 +1,505 @@
...
+    ``resolve_osm`` maps a scene token to its ``lanelet2_map.osm`` path, and each
+    scene's map is parsed once and reused. A dataset adapter supplies the
+    resolver (it knows the on-disk scene layout), tests inject a direct path.
+    """
+
+    def __init__(self, resolve_osm) -> None:
+        """Store the scene-token to OSM-path resolver.
+
+        Args:
+            resolve_osm: Callable mapping a scene token to its OSM path.
```
mojomex: Also needs an `.available()` method, so please document that too.

mojomex (commented):
Some small feedback, mostly LGTM. I'd prefer `lanelet2` lib over manual parsing as it's hard to keep in sync and validate edge cases.

autoware_ml/metrics/geometry/lanelet.py
amadeuszsz: You were right! Addressed in https://github.com/tier4/autoware-ml/pull/99/commits/cb0ac01b9b6716a8bbf3cd4072d805022c8665a5.

autoware_ml/metrics/geometry/lanelet.py
amadeuszsz: No, on our reference map it turned 2 rings into multi-polygons that were then passed on as polygons. make_valid runs now and every part is kept, unrepairable rings raise. Addressed in https://github.com/tier4/autoware-ml/pull/99/commits/66e85124e4392878aeee3b201276df3f0c7d8c7a.

autoware_ml/metrics/geometry/lanelet.py
amadeuszsz: Agreed, an unparsable or non-positive speed_limit now raises with the lanelet id and the raw value. A missing tag still falls back to the default. Addressed in https://github.com/tier4/autoware-ml/pull/99/commits/3090e0841d020515a1738a6bb68a5f04891b45ed.

autoware_ml/metrics/geometry/lanelet.py
amadeuszsz: Switched to max, a slower agent reaches less far and the model is worst case, so criticality shifts where lanelets overlap. Stacked lanelets read the same way since membership is 2D, now noted in the docstring. Addressed in https://github.com/tier4/autoware-ml/pull/99/commits/7a1e7334afab62ba88c5bc180a612d392922c84d.

autoware_ml/metrics/geometry/lanelet.py
amadeuszsz: Done, the cache moved onto the public method. Addressed in https://github.com/tier4/autoware-ml/pull/99/commits/6fdaa95367451b8e41e0163d00e1677231fdaad8.

autoware_ml/metrics/geometry/lanelet.py
amadeuszsz: Both membership tests reject negative margins now. Kept the two arguments though, erosion cuts against the outer border of the whole surface while expansion claims off-map area, so one signed number would hide that they are different geometries. Addressed in https://github.com/tier4/autoware-ml/pull/99/commits/34c9685d2bc9e3e6bda9a84a829c37c742cb666d.

autoware_ml/metrics/geometry/lanelet.py
amadeuszsz: Dropped it instead, from_osm already caches by path so the provider dict was a second cache in front of the first. Addressed in https://github.com/tier4/autoware-ml/pull/99/commits/a41295b64aea4e4f3f4a629e6a27b0418dbeae51.

autoware_ml/metrics/geometry/lanelet.py
amadeuszsz: Done, both calls are now an OsmPathResolver protocol. Addressed in https://github.com/tier4/autoware-ml/pull/99/commits/77152c8357922e3213824d496f392c0478134e0f.

autoware_ml/metrics/geometry/lanelet.py
amadeuszsz: On reflection your second point is the better design, so margin is now a single signed distance: negative erodes, positive dilates, and expand is gone. The key token spells the sign too, region_walkway_minus0p2 instead of region_walkway_m0.2. Addressed in https://github.com/tier4/autoware-ml/pull/99/commits/c42edf226ba7fca4adeabdefae830e00cdeac30b.

mojomex (approved):
LGTM! Seeing the code, I'm kinda sorry for proposing lanelet2 - I assumed it would cover all OSM features we are using. Your call if you're happier with the current version or the one before.

Tests are looking good!

PR #100, opened 2026-08-25
feat(metrics): worst-case reachability collision model

amadeuszsz:
## Summary

Worst-case reachability time-to-collision model for the upcoming criticality metrics (B1, B2).

- Planner-independent TTC: every agent, ego included, travels at its class max speed, and TTC is the first look-ahead time at which the two reachable-at-`t` sets overlap. Matched-speed lead traffic never meets ego, so its TTC is inf and its collision weight is 0.
- Three reachable-set kinds: wheeled (constant-curvature arc fronts under a friction-limited turn radius, clipped to the drivable area), VRU (isotropic disc), static (fixed footprint).
- Distance-bound prescreen resolves most objects to TTC = inf without stepping, and `EgoReachability` shares the per-frame ego geometry across all objects.
- Wheeled max speed comes from the lanelet `speed_limit` (#99), with a spec-versioned constant off-map. Parameters live in a frozen `ReachabilityParams` dataclass.

## Related Issues

Stacked on #99.

## Checklist

## Additional Notes

Purely additive module, nothing references it yet. The collision provider and the weighted metrics consuming it arrive with the criticality PR next in the stack.

## Additional Log and Media

`pytest autoware_ml/tests/{metrics,configs,datamodule}`: 139 passed at this branch head.

### DIFF
--- a/autoware_ml/metrics/detection3d/suite.py
+++ b/autoware_ml/metrics/detection3d/suite.py
@@ -3,7 +3,10 @@
 ``Detection3DMetricSuite`` owns the per-frame prediction and ground-truth tensors
 as list states (so torchmetrics handles cross-GPU sync) and applies the GT
 filters at ``update`` time. It knows nothing about which metrics run: it builds
-a ``DetectionState`` (overall and per range) and hands it to the injected metrics.
+a ``DetectionState`` (overall and per range/filter) and hands it to the injected
+metrics. Region and collision filters need per-frame metadata (ego pose, scene token) in the
+eval output, so their keep-masks are precomputed per frame at update and nothing needs the map
+at compute time, DDP gathers only tensors.
 """
 
 from __future__ import annotations
@@ -11,16 +14,41 @@ from __future__ import annotations
 import logging
 from typing import Any
 
+import numpy as np
 import torch
 
-from autoware_ml.metrics.base import Metric, MetricRange, MetricSuite
+from autoware_ml.metrics.base import (
+    IDENTITY,
+    Metric,
+    MetricFilter,
+    MetricRange,
+    MetricSuite,
+    rank_zero,
+)
+from autoware_ml.metrics.class_groups import fold_labels, resolve_class_groups
 from autoware_ml.metrics.detection3d.matching import (
+    DetectionState,
     clip_to_range,
     gt_keep_mask,
+    resolve_match_cost,
 )
-from autoware_ml.metrics.detection3d.structures import Detection3DSample, DetectionState
+from autoware_ml.metrics.detection3d.naming import metric_token
+from autoware_ml.metrics.detection3d.structures import Detection3DSample
 
 logger = logging.getLogger(__name__)
+
+
+def _validate_name_tokens(names: tuple[str, ...] | None) -> None:
+    """Reject class-name sets whose metric-key tokens collide."""
+    if not names:
+        return
+    tokens = [metric_token(name) for name in names]
+    duplicates = sorted({token for token in tokens if tokens.count(token) > 1})
+    if duplicates:
+        raise ValueError(
+            f"class names collapse to duplicate metric-key tokens {duplicates}, "
+            "per-class values would silently overwrite each other."
+        )
 
 
 class Detection3DMetricSuite(MetricSuite[DetectionState]):
@@ -36,7 +64,6 @@ class Detection3DMetricSuite(MetricSuite[DetectionState]):
         self,
         components: list[Metric[DetectionState]],
         class_names: tuple[str, ...] | None = None,
-        thresholds: tuple[float, ...] = (0.5, 1.0, 2.0, 4.0),
         ranges: tuple[MetricRange, ...] = (
             MetricRange("0-50m", 0.0, 50.0),
             MetricRange("50-90m", 50.0, 90.0),
@@ -45,13 +72,48 @@ class Detection3DMetricSuite(MetricSuite[DetectionState]):
         ),
         eval_class_range: dict[str, float] | None = None,
         min_num_points: int = 0,
+        match_cost: str = "center",
+        class_groups: dict[str, list[str]] | None = None,
+        collision: Any = None,
         **kwargs: Any,
     ) -> None:
+        """Validate the class setup and register the accumulation states.
+
+        Args:
+            components: Injected metrics, as in :class:`MetricSuite`.
+            class_names: Ordered class names for metric keys, or ``None`` to report only
+                classes present in the ground truth.
+            ranges: Radial windows every key is also emitted for.
+            eval_class_range: Class name to maximum GT distance in meters, applied as a GT
+                filter at accumulation.
+            min_num_points: Minimum LiDAR points a GT box needs to count, 0 disables the
+                filter.
+            match_cost: Matching cost name, ``center`` or ``corner``.
+            class_groups: Group name to member class names. When set, labels are folded onto
+                the group taxonomy at state build, so matching happens between grouped classes.
+            collision: Optional collision TTC provider. When set, the per-box
+                time-to-collision to ego is computed once per frame at update and carried on
+                each sample for the criticality metrics.
+            **kwargs: Forwarded to :class:`MetricSuite`.
+        """
         super().__init__(components=components, ranges=ranges, **kwargs)
         self.class_names = tuple(class_names) if class_names is not None else None
-        self.thresholds = tuple(float(threshold) for threshold in thresholds)
+        _validate_name_tokens(self.class_names)
         self.eval_class_range = eval_class_range
         self.min_num_points = int(min_num_points)
+        self.match_cost = str(match_cost)
+        resolve_match_cost(self.match_cost)  # a config typo must fail at build, not after the epoch
+        self.collision = collision
+        # When class_groups is set, GT/pred labels are folded onto the behaviour
+        # taxonomy at state build (after the per-class GT filters), so matching
+        # happens between grouped classes: a bus predicted as a truck is a true
+        # positive for grouped_vehicle. The synced boxes stay the trained classes.
+        self._group_lut, self._grouped_names = (
+            resolve_class_groups(self.class_names, class_groups) if class_groups else (None, None)
+        )
+        # The grouped names are what label_metric_name tokenizes in a grouped
+        # suite, so they need the same duplicate-token guard.
+        _validate_name_tokens(self._grouped_names)
 
         if eval_class_range and not self.class_names:
             raise ValueError("class_names must be provided when eval_class_range is configured.")
@@ -62,6 +124,66 @@ class Detection3DMetricSuite(MetricSuite[DetectionState]):
         self.add_state("pred_labels", default=[], dist_reduce_fx=None)
         self.add_state("gt_boxes", default=[], dist_reduce_fx=None)
         self.add_state("gt_labels", default=[], dist_reduce_fx=None)
+        if self.collision is not None:
+            self.add_state("gt_ttc", default=[], dist_reduce_fx=None)
+            self.add_state("pred_ttc", default=[], dist_reduce_fx=None)
+            # Per-frame flag: does this frame's scene have a lanelet map (and the
+            # batch an ego pose)? Uncovered frames carry all-inf TTC and are
+            # excluded from the criticality metrics' denominators, and the counters
+            # log the coverage once per epoch, like the region filters do.
+            self.add_state("ttc_covered", default=[], dist_reduce_fx=None)
+            self.add_state(
+                "ttc_frames_seen", default=torch.zeros((), dtype=torch.long), dist_reduce_fx="sum"
+            )
+            self.add_state(
+                "ttc_frames_covered",
+                default=torch.zeros((), dtype=torch.long),
+                dist_reduce_fx="sum",
+            )
+
+        # Distinct non-identity filters used by the components. Their keep-masks
+        # are precomputed per frame at update (where the ego pose and map are
+        # available) and stored as boolean tensors, so nothing needs the scene
+        # token or map at compute time and DDP gathers only tensors.
+        self._register_component_filters()
+        if self._component_filters:
+            self.add_state("gt_region_masks", default=[], dist_reduce_fx=None)
+            self.add_state("pred_region_masks", default=[], dist_reduce_fx=None)
+            self.add_state("region_available", default=[], dist_reduce_fx=None)
+
+    def compute(self) -> dict[str, float]:
+        """Log TTC frame coverage once per epoch, then run the metric groups.
+
+        Returns:
+            Metric keys mapped to values.
+        """
+        self._log_ttc_coverage()
+        return super().compute()
+
+    def _log_ttc_coverage(self) -> None:
+        """Log how many frames the collision provider covered (map + ego pose).
+
+        Uncovered frames are excluded from the criticality metrics' denominators
+        (the whole-scene metrics keep them), so a partial-map dataset is never
+        silently deflated.
+        """
+        if self.collision is None or not rank_zero():
+            return
+        seen = int(self.ttc_frames_seen.item())
+        covered = int(self.ttc_frames_covered.item())
+        if not seen:
+            return
+        if covered < seen:
+            logger.warning(
+                "Collision TTC covered %d/%d frames, %d frame(s) in scenes without a "
+                "lanelet map are excluded from the criticality metrics (whole-scene "
+                "metrics keep them).",
+                covered,
+                seen,
+                seen - covered,
+            )
+        else:
+            logger.info("Collision TTC covered %d/%d frames (full map coverage).", covered, seen)
 
     def _warn_on_range_class_caps(self) -> None:
         if self.eval_class_range is None:
@@ -79,8 +201,29 @@ class Detection3DMetricSuite(MetricSuite[DetectionState]):
                         metric_range.name,
                     )
 
+    def required_keys(self) -> tuple[str, ...]:
+        """Fold in the collision provider's frame-context keys.
+
+        The collision TTC needs the ego pose and scene token exactly like the
+        region filters do. Declaring them here makes a missing key fail loud at
+        the first batch instead of silently reporting zero coverage.
+
+        Returns:
+            The required eval-output key names.
+        """
+        keys = super().required_keys()
+        if self.collision is not None:
+            keys = tuple(dict.fromkeys(keys + ("ego2global", "scene_token")))
+        if self.min_num_points > 0:
+            keys = tuple(dict.fromkeys(keys + ("gt_num_points",)))
+        return keys
+
     def update(self, eval_out: dict[str, Any]) -> None:
-        """Accumulate one batch, applying every GT filter per frame."""
+        """Accumulate one batch, applying every GT filter per frame.
+
+        Args:
+            eval_out: Flat eval-output dict built by the model for one batch.
+        """
         predictions = eval_out["predictions"]
         gt_boxes = eval_out["gt_boxes"]
         gt_labels = eval_out["gt_labels"]
@@ -88,6 +231,11 @@ class Detection3DMetricSuite(MetricSuite[DetectionState]):
         if len(predictions) != len(gt_boxes) or len(predictions) != len(gt_labels):
             raise ValueError(
                 "Detection metric expects equal numbers of predictions, gt_boxes, and gt_labels."
+            )
+        if self.min_num_points > 0 and gt_num_points is None:
+            raise ValueError(
+                f"min_num_points={self.min_num_points} is configured but the eval output "
+                "carries no gt_num_points, the GT filter cannot be applied."
             )
 
         for i, (prediction, boxes, labels) in enumerate(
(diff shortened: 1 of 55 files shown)


autoware_ml/metrics/detection3d/suite.py
```diff
@@ -109,41 +251,213 @@ def update(self, eval_out: dict[str, Any]) -> None:
...
+                )
+
+            if self._component_filters and self._stage_filter_keys():
+                available = self._frame_region_availability(eval_out, i)
+                self.region_available.append(
+                    torch.tensor(available, dtype=torch.bool, device=frame_boxes.device)
+                )
+                self.region_available.append(
+                    torch.tensor(available, dtype=torch.bool, device=frame_boxes.device)
+                )
```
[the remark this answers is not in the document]
amadeuszsz: Right, the availability list was appended twice per frame, which desynchronizes it from the per-frame masks once coverage varies between scenes. Fixed in #98 where this state lives: a single append per frame remains.

autoware_ml/metrics/geometry/reachability.py
```diff
@@ -0,0 +1,349 @@
...
+            drivable: Road-region union in the map frame that wheeled fronts are clipped to.
+            params: Propagation parameters shared by every evaluation.
+        """
+        if ego.kind != WHEELED:
+            raise ValueError("ego must be a wheeled agent.")
+        if drivable is None:
+            raise ValueError("ego reachability needs a drivable polygon to clip against.")
+        self.ego = ego
+        self.params = params
+        self.steps = int(round(params.horizon_s / params.dt_s))
```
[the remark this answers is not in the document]
amadeuszsz: Fixed. The step count now floors with a small tolerance, so an exact multiple of dt_s keeps its final step and a non-divisible horizon never gains a step beyond it. Params also reject dt_s larger than horizon_s, and tests cover the 1.0/0.6 case.

autoware_ml/types/metrics.py
```diff
@@ -0,0 +1,16 @@
+from enum import StrEnum, auto
+
+
+class AgentKind(StrEnum):
+    """
+    Reachable-set shape of a collision participant.
+
+    Attributes:
+      WHEELED: Agent bound to a road surface, reaching along kinematically feasible arcs.
+      VRU: Agent free to move any direction, reaching an isotropic disc.
```
mojomex: nit: sometimes bicyclists are defined as VRUs too, so VRU might be a slightly unintuitive name

autoware_ml/metrics/geometry/reachability.py
```diff
@@ -0,0 +1,536 @@
...
+        t: Time offset in seconds.
+        params: Shared reachability parameters.
+        drivable: Drivable surface the arcs must stay on.
+
+    Returns:
+        The buffered front geometry, empty when no arc is feasible.
+    """
+    length = agent.speed * t
+    if length <= 1e-6:
+        return Point(agent.x, agent.y).buffer(agent.body_radius)
```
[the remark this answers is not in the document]
amadeuszsz: Right, the stopped branch skipped the surface clip, so a parked car could buffer across a median into ego's lane. It now goes through the same connected-sweep clipping as a moving front, and an off-surface stopped agent gets an empty body like an infeasible arc. Addressed in https://github.com/tier4/autoware-ml/pull/100/commits/58a1d0376bab45224a227d21e716579dbc38cd5f.

autoware_ml/metrics/geometry/reachability.py
```diff
@@ -0,0 +1,538 @@
...
+"""Reachability time-to-collision for the criticality metrics.
+
+One planner-independent model. Every agent, ego included, travels at the
+maximum map-legal speed for its class. The recorded path and observed
+velocities are never used. Time-to-collision is the earliest look-ahead time
+``t`` at which ego and the object can occupy a common point at the same time
+``t``, each having travelled exactly ``speed * t`` along a feasible path
+(reachable at ``t``, not "arrive early and wait"). That is what makes
+same-direction traffic at matched speed non-critical: the two fronts stay a
+constant gap apart and never meet, so TTC = inf and its collision weight is 0.
```
mojomex: This is my biggest issue with this PR: TTC for _perception_ should answer the question:

"In a scenario where the ego does not detect an agent, what is the worst-case consequence that can come of that"?

So, _given the worst-case action the other agent and ego take_, what is the time it takes for them to collide.

Consequently, agents immediately in front of the ego should have a TTC close to 0, not `inf`.

My suggestion is to aggregate reachable sets with all worst-case velocities: `+v_max` and `-v_max` into one reachable-at-t set.

autoware_ml/metrics/geometry/reachability.py
```diff
@@ -0,0 +1,538 @@
...
+
+Reachable-at-``t`` set by class, in the map frame (metres):
+
+* wheeled (car / truck / bus / train / motorcycle, and ego): the endpoints
+  of every feasible constant-curvature arc of length ``v * t`` under bounded
+  steering (minimum turn radius), a curved front, clipped to the drivable
+  area (an arc leaving the road is infeasible) and given the vehicle's body
+  half-width.
```
mojomex: Completely enough for now, but a more accurate motion model (piecewise arcs etc.) might be necessary to find the full reachable set in tight road geometries where one arc might leave the road temporarily.

autoware_ml/metrics/geometry/reachability.py
mojomex: In the future, having more realistic constraints, possibly based on GT velocity and `a_min`, `a_max` might be considered.

autoware_ml/metrics/geometry/reachability.py
```diff
@@ -0,0 +1,538 @@
...
+import shapely
+from shapely.geometry import LineString, Point, Polygon, box
+from shapely.geometry.base import BaseGeometry
+
+from autoware_ml.types.metrics import AgentKind
+
+# Abutting map ways are authored with independently rounded coordinates, so their
+# union can keep a hairline seam. An exact path check would reject every arc that
+# crosses one, so a surface is sealed by this much before it is used.
+SURFACE_TOLERANCE_M = 1e-6
```
mojomex: Realistically, even millimeter or centimeter precision (`1e-3` or `1e-2`) would be completely sufficient. Feel free to relax it.

autoware_ml/metrics/geometry/reachability.py
```diff
@@ -0,0 +1,538 @@
...
+
+# The reachable region's far edge is a chord approximation of the far arc, so a
+# front can poke a fraction of a millimetre past it. The prescreen bound is
+# inflated by this much to stay a true superset of every front.
+REGION_SLACK_M = 1e-3
```
mojomex: Same here.

autoware_ml/metrics/geometry/reachability.py
```diff
@@ -0,0 +1,538 @@
...
+    half-extent added so a collision is a footprint overlap, not a point
+    coincidence.
+    """
+
+    kind: AgentKind
+    x: float
+    y: float
+    heading: float = 0.0
+    speed: float = 0.0
+    body_radius: float = 0.5
```
mojomex: Initializing heading and other params to defaults might silently produce wrong results when someone forgets to set it - better not have a default.

For agents that have isotropic shape and movement (e.g. pedestrians), these values can be defaulted, but not for vehicles.

autoware_ml/metrics/geometry/reachability.py
```diff
@@ -0,0 +1,538 @@
...
+        drivable: Drivable surface the arcs must stay on.
+
+    Returns:
+        The buffered front geometry, empty when no arc is feasible.
+    """
+    length = agent.speed * t
+    if length <= 1e-6:
+        # A stopped agent still owns its body, clipped like a moving front so it
+        # cannot spill over a kerb or a median into a lane it cannot drive to.
+        return _connected_sweep(Point(agent.x, agent.y), agent.body_radius, drivable)
```
mojomex: Are all agents assumed to be circular? The error term for vehicles, especially trucks, would be huge 👀

autoware_ml/metrics/geometry/reachability.py
```diff
@@ -0,0 +1,538 @@
...
+    kmax = 1.0 / params.turn_radius(agent.speed)
+    grid = _arc_grid(agent, length, kmax, params.arc_samples)
+    runs: list[list[tuple[float, float]]] = [[]]
+    for arc in grid:
+        if drivable.covers(LineString(arc)):
+            runs[-1].append((float(arc[-1][0]), float(arc[-1][1])))
+        elif runs[-1]:
+            runs.append([])
```
mojomex: Does this serve any function?

autoware_ml/metrics/geometry/reachability.py
```diff
@@ -0,0 +1,538 @@
...
+        The buffered front geometry, empty when no arc is feasible.
+    """
+    length = agent.speed * t
+    if length <= 1e-6:
+        # A stopped agent still owns its body, clipped like a moving front so it
+        # cannot spill over a kerb or a median into a lane it cannot drive to.
+        return _connected_sweep(Point(agent.x, agent.y), agent.body_radius, drivable)
+    kmax = 1.0 / params.turn_radius(agent.speed)
+    grid = _arc_grid(agent, length, kmax, params.arc_samples)
+    runs: list[list[tuple[float, float]]] = [[]]
```
mojomex: Please have a one-line comment describing the contents of ``runs``

autoware_ml/metrics/geometry/reachability.py
```diff
@@ -0,0 +1,538 @@
...
+        )
+        for run in runs
+        if run
+    ]
+    if not pieces:
+        return Polygon()
+    return shapely.union_all(pieces)
+
+
+def reachable_region(
```
mojomex: Docstring says this is for wheeled agent, so please make that clear in the function name (similar to `wheeled_front`)

autoware_ml/metrics/geometry/reachability.py
```diff
@@ -0,0 +1,538 @@
...
+        agent: Wheeled agent in the map frame.
+        params: Shared reachability parameters.
+        drivable: Drivable surface the region is clipped to.
+
+    Returns:
+        The filled reachable region.
+    """
+    length = agent.speed * params.horizon_s
+    if length <= 1e-6:
+        region = Point(agent.x, agent.y).buffer(agent.body_radius)
```
mojomex: Same body shape comment as above

autoware_ml/metrics/geometry/reachability.py
```diff
@@ -0,0 +1,538 @@
...
+
+    A clipped sweep can span several disconnected drivable polygons, and only
+    the part road-connected to the agent is truly reachable.
+    """
+    parts = _polygonal_parts(region)
+    if not parts:
+        return Polygon()
+    if len(parts) == 1:
+        return parts[0]
+    return min(parts, key=seed.distance)
```
mojomex: Not having this extra `len == 1` check should come at no measurable const and is simpler. Best remove.

autoware_ml/metrics/geometry/reachability.py
```diff
@@ -0,0 +1,538 @@
...
+            # empty surface on purpose: nothing is reachable, every object scores
+            # zero risk, and the frame still contributes to the other metrics.
+            return inf
+        ego, params = self.ego, self.params
+
+        if obj.kind == AgentKind.STATIC:
+            # A static set never grows: it must already meet the hat.
+            if not self._hat.intersects(obj.footprint):
+                return inf
+            # Ego's front is within speed * t + body of ego's position.
```
mojomex: ~Ego's front~ -> _Agent's front_?

autoware_ml/tests/metrics/test_reachability.py
```diff
@@ -0,0 +1,269 @@
...
+PARAMS = ReachabilityParams(horizon_s=4.0, dt_s=0.1)
+# A wide open drivable region so wheeled fronts are not clipped in these unit cases.
+ROAD = box(-80.0, -50.0, 500.0, 50.0)
+
+
+def _footprint(x: float, y: float, size: float = 2.0):
+    return box(x - size / 2, y - size / 2, x + size / 2, y + size / 2)
+
+
+def test_same_speed_lead_never_collides() -> None:
```
mojomex: I'd explicitly request the opposite, as commented above.

autoware_ml/tests/metrics/test_reachability.py
```diff
@@ -0,0 +1,269 @@
...
+    # Following a vehicle at matched speed: fronts stay a constant gap apart -> inf.
+    ego = Agent(AgentKind.WHEELED, 0.0, 0.0, heading=0.0, speed=10.0, body_radius=1.2)
+    lead = Agent(AgentKind.WHEELED, 25.0, 0.0, heading=0.0, speed=10.0, body_radius=1.2)
+    assert time_to_collision(ego, lead, ROAD, PARAMS) == inf
+
+
+def test_stationary_object_ahead_finite_near_distance_over_speed() -> None:
```
mojomex: Nit: Test name is a bit unintuitive.

autoware_ml/tests/metrics/test_reachability.py
```diff
@@ -0,0 +1,269 @@
...
+
+
+def test_crossing_vru_is_finite_within_horizon() -> None:
+    ego = Agent(AgentKind.WHEELED, 0.0, 0.0, heading=0.0, speed=10.0, body_radius=1.0)
+    ped = Agent(AgentKind.VRU, 18.0, 6.0, speed=4.0, body_radius=0.4)
+    ttc = time_to_collision(ego, ped, ROAD, PARAMS)
+    assert ttc != inf and ttc <= PARAMS.horizon_s
+
+
+def test_far_object_is_rejected_cheaply() -> None:
```
mojomex: The "cheaply" part can't really be verified here, please consider renaming.

autoware_ml/tests/metrics/test_reachability.py
```diff
@@ -0,0 +1,269 @@
...
+            obj = Agent(AgentKind.VRU, x, y, speed=float(rng.uniform(0.0, 6.0)), body_radius=0.4)
+        else:
+            obj = Agent(
+                AgentKind.WHEELED,
+                x,
+                y,
+                heading=float(rng.uniform(0, 2 * pi)),
+                speed=float(rng.uniform(0.5, 16.7)),
+                body_radius=1.0,
+            )
```
mojomex: For readability, best factor this out into a helper function.

autoware_ml/tests/metrics/test_reachability.py
mojomex, line 1: One more request: an L-shaped one-lane road (e.g. a small intersection) where the ego approaches from one leg, and a wheeled agent from the other.

They should always have a finite TTC (e.g. when both driving straight).
This is to test that the simplified arc motion model is not overly pessimistic in complex geometries.

mojomex (commented):
Thanks. Most feedback is minor, but the "vehicle in front of ego has infinite TTC" rule seems overly dangerous to me - in real life, we have to expect that any agent might slam the brakes at any moment and leave enough space behind them. This necessitates that we detect them in the first place. Infinite TTC would basically say "it's fine (weight 0.0) not to detect the agent in front of us", which I don't agree with.

autoware_ml/types/metrics.py
amadeuszsz: Fair point, cyclists are VRUs in most definitions and they are wheeled here. The kind is now `living`, which says what it means without borrowing a term. Addressed in https://github.com/tier4/autoware-ml/pull/100/commits/be4f27ab04820bf6ac7b7aa8826512223553b0c3.

autoware_ml/metrics/geometry/reachability.py
amadeuszsz: Agreed, and this was the right thing to push back on. Every agent now sweeps up to `v_max * t` forward or in reverse, whichever closes the gap, so a close lead has a small TTC instead of inf. The module docstring states the perception framing you describe. Addressed in https://github.com/tier4/autoware-ml/pull/100/commits/9847b859bc33964b55091fc2d6bbc9302f12b63f.

autoware_ml/metrics/geometry/reachability.py
amadeuszsz: Noted as a known limitation: one arc per curvature, truncated where it leaves the road, so a lane that needs two arcs is under-reached. I added a one-lane junction test to pin down that the simple model still finds the obvious collisions, see https://github.com/tier4/autoware-ml/pull/100/commits/063ea5244e62be16fb130a2439b4ad322b0d6c59.

autoware_ml/metrics/geometry/reachability.py
amadeuszsz: Relaxed to 1 mm for the surface seal. Addressed in https://github.com/tier4/autoware-ml/pull/100/commits/f854c97aab4a549c5ec803c38fd7a63fab2a3757.

autoware_ml/metrics/geometry/reachability.py
amadeuszsz: Relaxed to 1 cm for the prescreen margin, same commit: https://github.com/tier4/autoware-ml/pull/100/commits/f854c97aab4a549c5ec803c38fd7a63fab2a3757.

autoware_ml/metrics/geometry/reachability.py
amadeuszsz: Agreed. Nothing is defaulted any more, and agents are built through `Agent.wheeled`, `Agent.living` and `Agent.static`, so an isotropic agent never has to name a heading and a vehicle cannot forget one. Addressed in https://github.com/tier4/autoware-ml/pull/100/commits/2cf13c701310f5b881daad05ee28494a98a5326b.

autoware_ml/metrics/geometry/reachability.py
amadeuszsz: They were, and you are right about the error. A wheeled agent now sweeps its box: the driven path is extended by half the body length and buffered by half the width, so a truck reaches metres further than a disc would. Objects take their extents from the detection box, ego from config. Addressed in https://github.com/tier4/autoware-ml/pull/100/commits/672480e00ab3de3a5c696b084388a896c89b7664.

autoware_ml/metrics/geometry/reachability.py
amadeuszsz: It split the front where a curvature sector was infeasible, so the endpoint locus was not bridged across it. The rework drops the bookkeeping: each arc is truncated and swept on its own and the union is the set. Addressed in https://github.com/tier4/autoware-ml/pull/100/commits/9847b859bc33964b55091fc2d6bbc9302f12b63f.

autoware_ml/metrics/geometry/reachability.py
amadeuszsz: `runs` is gone with the rework, each arc is now truncated and swept independently. Addressed in https://github.com/tier4/autoware-ml/pull/100/commits/9847b859bc33964b55091fc2d6bbc9302f12b63f.

autoware_ml/metrics/geometry/reachability.py
amadeuszsz: Renamed to `wheeled_reachable_region`, and it is now a thin call into `wheeled_reachable_set` at the horizon so the hat and the steps cannot drift apart. Addressed in https://github.com/tier4/autoware-ml/pull/100/commits/9847b859bc33964b55091fc2d6bbc9302f12b63f.

autoware_ml/metrics/geometry/reachability.py
amadeuszsz: Same fix, the region is the same swept-body call at the horizon. Addressed in https://github.com/tier4/autoware-ml/pull/100/commits/672480e00ab3de3a5c696b084388a896c89b7664.

autoware_ml/metrics/geometry/reachability.py
amadeuszsz: Removed. Addressed in https://github.com/tier4/autoware-ml/pull/100/commits/ec6b4332948f63031d3fd16155afb05f52bcab19.

autoware_ml/metrics/geometry/reachability.py
amadeuszsz: Reworded. The line is about ego (only ego moves against a static set), which is what made it confusing, so it now says that outright. Addressed in https://github.com/tier4/autoware-ml/pull/100/commits/d646bd18143e6b452320c99700ec6a27aeb6c6e9.

autoware_ml/tests/metrics/test_reachability.py
amadeuszsz: Done, the test now asserts a finite TTC of about 0.8 s for a 25 m lead at matched speed, with the reasoning in the comment. Addressed in https://github.com/tier4/autoware-ml/pull/100/commits/9847b859bc33964b55091fc2d6bbc9302f12b63f.

autoware_ml/tests/metrics/test_reachability.py
amadeuszsz: Renamed to `test_a_stationary_object_ahead_collides_at_about_distance_over_speed`. Addressed in https://github.com/tier4/autoware-ml/pull/100/commits/b7594378680328dd145bf832a087fc2afc90b957.

autoware_ml/tests/metrics/test_reachability.py
amadeuszsz: True, it only checks the answer. Renamed to `test_a_far_object_is_unreachable_within_the_horizon`, same commit: https://github.com/tier4/autoware-ml/pull/100/commits/b7594378680328dd145bf832a087fc2afc90b957.

autoware_ml/tests/metrics/test_reachability.py
amadeuszsz: Factored into a `_sampled_agent` helper. Addressed in https://github.com/tier4/autoware-ml/pull/100/commits/ba1dab94b8cc497b1e1f43254104430a427625d0.

autoware_ml/tests/metrics/test_reachability.py
amadeuszsz: Added: two legs of a one-lane bend, ego on one and a wheeled agent on the other, both driving straight, asserting a finite TTC around 2.7 s. Addressed in https://github.com/tier4/autoware-ml/pull/100/commits/063ea5244e62be16fb130a2439b4ad322b0d6c59.

autoware_ml/metrics/geometry/reachability.py
```diff
@@ -0,0 +1,724 @@
...
+
+    The agent may travel up to ``speed * t`` forward or in reverse along any
+    constant-curvature arc its minimum turn radius allows, so the set is the body
+    swept over every such arc, truncated where an arc leaves the road, clipped to
+    the surface, and reduced to the part still connected to the agent (a body
+    must not hop a kerb into a lane it cannot drive to). An agent whose reference
+    point is off the surface reaches nothing.
+
+    The set grows with ``t``, so the same call at the horizon is the region the
+    prescreen and the collision filter use.
```
mojomex: This no longer holds since the body sweep: the shorter arc's straight tangent extension (by `front`) pokes outside the full arc's curved capsule, so per-step sets escape the hat and the prescreen prunes real collisions.

Example on open road, ego `front=3.8, rear=1.1, width=2.0`: at 10 m/s step 37 escapes the hat by 5.8 m²; at 3 m/s the union of escapes is 25 m², and 9 of 24 small static obstacles placed there get `inf` from `EgoReachability` while brute-force stepping returns ~3 s. Even with the test's own 2.4/2.4/2.4 body, steps 36-39 escape (0.68 m² at 39) - `test_the_region_covers_every_step_including_the_first` only checks steps 1, 5, 10, 40.

Suggest building the hat as the union of the per-step sets (they are computed per frame anyway), and having the superset test iterate every step with an asymmetric body plus a low-speed ego in the brute-force test.

autoware_ml/metrics/geometry/reachability.py
```diff
@@ -0,0 +1,724 @@
...
+    pieces = []
+    # Forward and reverse: the worst case is whichever direction closes the gap.
+    for signed_reach in (reach, -reach):
+        for arc in _arc_grid(agent, signed_reach, kmax, params.arc_samples):
+            drivable_arc = _feasible_arc(LineString(arc), seed, drivable)
+            if drivable_arc is not None:
+                pieces.append(_body_sweep(drivable_arc, agent, drivable, signed_reach > 0.0))
+    if not pieces:
+        return Polygon()
+    return shapely.union_all(pieces)
```
mojomex, line 495: Future work (not blocking): with 21 curvature samples, adjacent arc endpoints at the horizon are ~2.3 m apart, more than a 2.0 m body width, so the outer edge of the union is scalloped and has uncovered notches in the last few tenths of a second of reach. The old endpoint-locus polygon bridged those. Either scale `arc_samples` with reach or bridge adjacent sweeps; please leave a note in the docstring or an issue.

autoware_ml/metrics/geometry/reachability.py
```diff
@@ -0,0 +1,724 @@
...
+    return min(parts, key=seed.distance)
+
+
+def reachable_set(
+    agent: Agent, t: float, params: ReachabilityParams, drivable: BaseGeometry | None
+) -> BaseGeometry:
+    """The agent's reachable-at-``t`` set in the map frame.
+
+    Wheeled fronts keep only arcs that stay on ``drivable`` for their whole path,
+    so a wheeled evaluation needs a drivable polygon. Living discs and static
```
mojomex: Please update: arcs are now truncated where they leave the surface, not discarded.

mojomex (dismissed):
Almost there. There's one real bug now that the body shape is not circular anymore, please have a look. Other feedback is future work.

autoware_ml/metrics/geometry/reachability.py
amadeuszsz: Confirmed, and the cause sits deeper than the nose extension: the strip around the axle path was only right at its two ends, the outer front corner of a turning bus runs up to 1.9 m outside it. The sweep is now the body rectangle placed at one pose per time step along the arc, so the corners are covered through the turn and a shorter travel's poses are a prefix of a longer one's. The horizon set is a superset again by construction. The superset test now runs over all 40 steps with the bus body, a 3 m/s ego and a narrow lane, and the brute-force test has the same egos. Cost: the hat goes from 10 ms to about 60 ms, a full 40-step frame from 0.3 s to about 1.3 s.

Addressed in https://github.com/tier4/autoware-ml/pull/100/commits/d15ec964ce651a299b5d3bc23cb63458190aaabb.

autoware_ml/metrics/geometry/reachability.py
amadeuszsz: Went with scaling: the curvature count is now derived from the horizon reach and the body so the far ends of neighbouring sweeps overlap (about 27 arcs for the bus at speed, 43 at the radius floor, 63 for a motorcycle), with arc_samples as the floor. What remains are shallow notches between the end faces, under a quarter of the body width deep, which the new test bounds against a 401-arc reference.

Addressed in https://github.com/tier4/autoware-ml/pull/100/commits/e5840eb3daf530357d4fc1ca0fcdb118fa483bb0.

autoware_ml/metrics/geometry/reachability.py
amadeuszsz: Addressed in https://github.com/tier4/autoware-ml/pull/100/commits/18dec8999cd964074bbc891e3c2a2749c9ac986f.

PR #101, opened 2026-08-25
feat(metrics): region, corridor and collision evaluation filters

amadeuszsz:
## Summary

Selection filters implementing the region, corridor and collision evaluation slices.

- `RegionFilter` keeps elements inside a union of lanelet2 regions, box footprints by overlap and points by membership, with an explicit `available()` check so map-less scenes are excluded from filtered slices instead of failing.
- `CorridorFilter` keeps a straight forward strip in the ego frame, map-free so the slice covers every scene, with distance left to the range axis.
- `CollisionFilter` is the filter form of the collision model (#100): it keeps whatever meets the area ego can reach within the horizon at max map-legal speed under bounded steering, clipped to the road lanelets.
- Filters declare the frame metadata they require and carry a stable name and cache key for suite slicing.

## Related Issues

Stacked on #100.

## Checklist

## Additional Notes

The detection suite consumes these through the filter axis added in #98. Segmentation suites follow later in the stack.

## Additional Log and Media

`pytest autoware_ml/tests/{metrics,configs,datamodule}`: 159 passed, 1 skipped at this branch head.

### DIFF
--- a/autoware_ml/metrics/filters.py
+++ b/autoware_ml/metrics/filters.py
@@ -1,0 +1,454 @@
+"""Concrete evaluation filters.
+
+The base :class:`~autoware_ml.metrics.base.MetricFilter` / ``IdentityFilter`` live
+in ``base.py`` (no heavy deps). Three filters live here. ``CorridorFilter`` keeps
+the elements inside a straight fixed strip ahead of ego and needs no map at all.
+``RegionFilter`` keeps the elements whose base_link position, transformed to the
+map frame by the per-frame ego pose, falls inside a chosen set of lanelet2
+regions, so any metric can be reported on the road or on the walkway.
+``CollisionFilter`` keeps the elements inside the ego collision area clipped to
+the road lanelets, the filter form of the collision model.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from functools import lru_cache
+from math import atan2
+from typing import Any
+
+import numpy as np
+import shapely
+import torch
+
+from autoware_ml.metrics.base import MetricFilter, number_token
+from autoware_ml.metrics.detection3d.geometry import bev_corners_batch
+from autoware_ml.metrics.geometry.lanelet import (
+    KNOWN_REGION_TOKENS,
+    LaneletMap,
+    LaneletMapProvider,
+)
+from autoware_ml.metrics.geometry.reachability import (
+    Agent,
+    ReachabilityParams,
+    VehicleGeometry,
+    localized_surface,
+    wheeled_reachable_region,
+)
+
+
+# The lanelet2 regions a vehicle may drive on, shared by every consumer of the
+# collision model so the filter and the TTC provider cannot drift apart.
+DEFAULT_DRIVABLE_REGION: tuple[str, ...] = ("road", "road_shoulder", "crosswalk")
+
+
+def as_numpy(value: Any) -> np.ndarray:
+    """Frame metadata as a NumPy array, from any device.
+
+    Collation and Lightning's device transfer turn per-frame metadata (the ego
+    pose) into tensors that live on the evaluation device, and ``np.asarray``
+    alone cannot read a CUDA tensor.
+
+    Args:
+        value: Array-like or tensor, possibly on a CUDA device.
+
+    Returns:
+        A NumPy array on the CPU.
+    """
+    if isinstance(value, torch.Tensor):
+        return value.detach().cpu().numpy()
+    return np.asarray(value)
+
+
+def _to_map_xy(xyz: np.ndarray, ego2global: Any) -> np.ndarray:
+    """Transform base_link ``xyz`` (N, 3+) to map-frame xy via ``ego2global``."""
+    transform = as_numpy(ego2global).astype(np.float64)
+    homogeneous = np.concatenate([xyz[:, :3], np.ones((xyz.shape[0], 1))], axis=1)
+    return (homogeneous @ transform.T)[:, :2]
+
+
+def ego2global_pose(ego2global: Any) -> tuple[float, float, float]:
+    """Ego map-frame ``(x, y, heading)`` from a 4x4 ego-to-global transform.
+
+    Args:
+        ego2global: 4x4 ego-to-global transform.
+
+    Returns:
+        The map-frame ``(x, y, heading)`` triple.
+    """
+    matrix = as_numpy(ego2global).astype(np.float64)
+    return float(matrix[0, 3]), float(matrix[1, 3]), atan2(float(matrix[1, 0]), float(matrix[0, 0]))
+
+
+def box_footprints_map(boxes: np.ndarray, ego2global: Any) -> np.ndarray:
+    """Map-frame BEV footprint polygons for detection boxes.
+
+    ``boxes`` rows are ``[cx, cy, cz, dx, dy, dz, yaw, ...]`` in base_link. Each
+    footprint's four corners are transformed to the map frame so region
+    membership can be tested by overlap rather than by the box center alone.
+
+    Args:
+        boxes: Box rows ``(N, 7+)`` in base_link.
+        ego2global: 4x4 ego-to-global transform of the frame.
+
+    Returns:
+        One shapely footprint polygon per box, in the map frame.
+    """
+    corners = bev_corners_batch(boxes)  # (N, 4, 2) base_link
+    heights = np.repeat(boxes[:, 2], corners.shape[1])
+    corners3 = np.column_stack([corners.reshape(-1, 2), heights])
+    map_xy = _to_map_xy(corners3, ego2global).reshape(corners.shape)
+    return shapely.polygons(map_xy)
+
+
+def _is_boxes(elements: np.ndarray) -> bool:
+    """Elements are detection boxes ``[cx,cy,cz,dx,dy,dz,yaw,...]`` rather than bare points."""
+    return elements.ndim == 2 and elements.shape[1] >= 7
+
+
+def ego_collision_agent(
+    lanelet_map: LaneletMap,
+    x: float,
+    y: float,
+    heading: float,
+    max_speed_mps: float,
+    vehicle: VehicleGeometry,
+) -> Agent:
+    """Builds a wheeled agent representing the ego for use in collision
+    calculations.
+
+    The collision area trim (:class:`CollisionFilter`) and the collision TTC
+    provider build ego through this single function, so the two can never drift:
+    the speed is the lanelet speed limit at ego's position with ``max_speed_mps``
+    as the off-map fallback. Callers hold the ego pose in different forms, so
+    they read it out of their transform with :func:`ego2global_pose` first.
+
+    Args:
+        lanelet_map: The scene's parsed lanelet map.
+        x: Ego map-frame x in meters.
+        y: Ego map-frame y in meters.
+        heading: Ego map-frame heading in radians.
+        max_speed_mps: Off-map fallback speed in m/s.
+        vehicle: Ego body dimensions.
+
+    Returns:
+        Ego as a wheeled agent in the map frame.
+    """
+    return Agent.wheeled(
+        x,
+        y,
+        heading,
+        lanelet_map.speed_at(x, y, max_speed_mps),
+        front=vehicle.front,
+        rear=vehicle.rear,
+        width=vehicle.width,
+        min_turn_radius=vehicle.min_turn_radius,
+    )
+
+
+def signed_number_token(value: float) -> str:
+    """Key-safe token spelling the sign, e.g. ``plus0p2`` or ``minus0p2``.
+
+    Zero has no direction, so it carries no sign word.
+
+    Args:
+        value: Signed number to encode.
+
+    Returns:
+        The token, sign word first for a non-zero value.
+    """
+    if value == 0.0:
+        return number_token(0.0)
+    return ("minus" if value < 0.0 else "plus") + number_token(abs(value))
+
+
+def validated_region(region: list[str] | tuple[str, ...], owner: str) -> tuple[str, ...]:
+    """The region tokens as a tuple, rejecting anything lanelet2 does not name.
+
+    A typo would otherwise surface as a silently empty slice, so both map-backed
+    filters validate through here.
+
+    Args:
+        region: Literal lanelet2 tokens.
+        owner: Name of the calling filter, for the error message.
+
+    Returns:
+        The tokens as a tuple of strings.
+
+    Raises:
+        ValueError: If ``region`` is empty or names an unknown token.
+    """
+    if not region:
+        raise ValueError(f"{owner} needs at least one lanelet2 region token.")
+    tokens = tuple(str(token) for token in region)
+    unknown = sorted(set(tokens) - KNOWN_REGION_TOKENS)
+    if unknown:
+        raise ValueError(
+            f"Unknown lanelet2 region tokens {unknown}, known tokens: "
+            f"{sorted(KNOWN_REGION_TOKENS)}. (A known token absent from a "
+            "particular scene's map is fine, that slice is simply empty.)"
+        )
+    return tokens
+
+
+@dataclass(frozen=True)
+class RegionFilter(MetricFilter):
+    """Keep only elements inside a union of lanelet2 regions.
+
+    Points near the outer border of the whole mapped surface are naturally
+    noisy (high entropy, frequent misclassification), so the border can be
+    moved by a signed margin, negative eroding it and positive dilating it.
+
+    Attributes:
+        region: List of literal lanelet2 tokens (lanelet ``subtype`` or area
+            ``type``), e.g. ``[road, road_shoulder, crosswalk]`` for the
+            drivable region or ``[walkway]`` for the pedestrian-only region.
+        map_provider: Resolves a scene token to its :class:`LaneletMap`.
+        margin: Border shift in meters. Negative erodes, so outer-border points
+            stop counting while internal borders between adjacent regions stay
+            intact. Positive dilates, so the region additionally claims off-map
+            points within the margin, never points of another mapped region.
+        name: Display name prefixing the metric keys. Derived from the tokens
+            and margin when left empty.
+    """
+
+    region: tuple[str, ...]
+    map_provider: LaneletMapProvider
+    margin: float = 0.0
+    name: str = ""
+
+    required_eval_keys = ("ego2global", "scene_token")
+
+    def __post_init__(self) -> None:
+        """Validate the region tokens and derive the display name."""
+        object.__setattr__(self, "region", validated_region(self.region, "RegionFilter"))
+        object.__setattr__(self, "margin", float(self.margin))
+        if not self.name:
+            margin_token = f"_{signed_number_token(self.margin)}" if self.margin else ""
+            object.__setattr__(self, "name", "region_" + "_".join(self.region) + margin_token)
+
+    @property
+    def cache_key(self) -> str:
+        """Every parameter that shapes the mask, equal keys must mean equal masks.
+
+        The provider is part of that: a suite keeps one mask set per key, so two
+        filters that read different maps must not share it.
+        """
+        region = ",".join(sorted(self.region))
+        return f"region:{region}:{self.margin:g}:{self.map_provider.cache_key}"
+
+    def keep(self, xyz: np.ndarray, context: dict[str, Any]) -> np.ndarray:
+        """Mask of elements in the region.
+
+        Detection boxes (7 or more columns) belong to the region when their
+        footprint overlaps it (any part inside), so an object overhanging from
+        an off-region center still counts. Segmentation points (3 columns) use
+        point-in-polygon.
+
+        Args:
+            xyz: Points ``(N, 3)`` or box rows ``(N, 7+)`` in base_link.
+            context: Per-frame values with the ego pose and scene token.
+
+        Returns:
+            Boolean mask of elements in the region.
+        """
+        xyz = np.asarray(xyz, dtype=np.float64)
+        if xyz.shape[0] == 0:
+            return np.zeros((0,), dtype=bool)
+        lanelet_map = self.map_provider.get(context["scene_token"])
+        if _is_boxes(xyz):
+            footprints = box_footprints_map(xyz, context["ego2global"])
+            return lanelet_map.intersects(self.region, footprints, self.margin)
+        map_xy = _to_map_xy(xyz, context["ego2global"])
+        return lanelet_map.contains(self.region, map_xy, self.margin)
+
+    def available(self, context: dict[str, Any]) -> bool:
+        """False for scenes with no lanelet map, so the suite excludes them.
+
+        Args:
+            context: Per-frame values with the scene token.
+
+        Returns:
+            Whether the scene has a lanelet map.
+        """
+        return self.map_provider.available(context["scene_token"])
+
+
+@dataclass(frozen=True)
+class CorridorFilter(MetricFilter):
+    """Keep only elements inside a straight corridor ahead of ego.
+
+    The corridor is a forward strip in the ego frame: ``width_m`` across,
+    centered on the x axis, with no length bound of its own because distance
+    slicing is the range axis's job. It needs no map and no pose, so the slice
+    covers every scene. A detection box is kept when its footprint overlaps
+    the strip, a segmentation point when it lies inside it.
+
+    Attributes:
+        width_m: Full corridor width in meters, centered on the x axis.
+        name: Display name prefixing the metric keys.
+    """
+
+    width_m: float = 3.0
+    name: str = "corridor"
+
+    required_eval_keys = ()
+
+    def __post_init__(self) -> None:
+        """Validate the strip width."""
+        if self.width_m <= 0.0:
+            raise ValueError("width_m must be > 0.")
+        object.__setattr__(self, "width_m", float(self.width_m))
+
+    @property
+    def cache_key(self) -> str:
+        """Every parameter that shapes the mask, equal keys must mean equal masks."""
+        return f"corridor:w{self.width_m:g}"
+
+    def keep(self, xyz: np.ndarray, context: dict[str, Any]) -> np.ndarray:
+        """Mask of elements inside the strip (footprint overlap / point test).
+
+        Args:
+            xyz: Points ``(N, 3)`` or box rows ``(N, 7+)`` in base_link.
+            context: Unused, the strip lives in the ego frame.
+
+        Returns:
+            Boolean mask of elements inside the strip.
+        """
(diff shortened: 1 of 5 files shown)


autoware_ml/metrics/filters.py
```diff
@@ -0,0 +1,360 @@
...
+
+class RegionFilter(MetricFilter):
+    """Keep elements inside a union of lanelet2 regions.
+
+    Points near the outer border of the whole mapped surface are naturally
+    noisy (high entropy, frequent misclassification), so the border can be
+    adjusted by a margin that either cuts inward or grows outward.
+    """
+
+    required_eval_keys = ("ego2global", "scene_token")
```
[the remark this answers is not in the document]
amadeuszsz: This is the stacking order, not a gap: the datamodules and `detection_eval_output` gain `ego2global` and `scene_token` in a follow-up PR of this stack, and no shipped config wires a map-backed filter or TTC provider until the config PR that comes after that wiring. At this branch head the map-aware path is present but not reachable from any built-in configuration.

autoware_ml/metrics/segmentation3d/suite.py
```diff
@@ -80,7 +80,9 @@ def update(self, eval_out: dict[str, torch.Tensor]) -> None:
                 device
             )
 
-    def state_for(self, metric_range: MetricRange | None) -> ConfusionState:
+    def state_for(
+        self, metric_range: MetricRange | None, metric_filter: MetricFilter = IDENTITY
+    ) -> ConfusionState:
```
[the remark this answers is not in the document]
amadeuszsz: Valid, the interim suite silently discarded the filter. It now rejects any component carrying a non-identity filter at construction, so a misconfiguration fails loud instead of emitting whole-scene values under a slice label. The successor suite later in this stack accumulates real per-filter buckets.

autoware_ml/metrics/geometry/reachability.py
```diff
@@ -0,0 +1,353 @@
...
+    so a wheeled evaluation needs a drivable polygon. VRU discs and static
+    footprints ignore ``drivable``.
+    """
+    if agent.kind == STATIC:
+        return agent.footprint
+    if agent.kind == VRU:
+        return Point(agent.x, agent.y).buffer(agent.speed * t + agent.body_radius)
+    front = wheeled_front(agent, t, params)
+    if drivable is not None:
+        front = front.intersection(drivable)
```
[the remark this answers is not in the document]
amadeuszsz: Fixed in #100 where the model lives. Each constant-curvature arc is now sampled as a polyline and kept only when the whole path stays on the drivable surface, so endpoints across a gap are unreachable and an off-surface wheeled agent has an empty front. The filled region used by the collision filter additionally keeps only the part connected to the agent. Covered by a disconnected-two-strips regression test.

autoware_ml/metrics/filters.py
```diff
@@ -0,0 +1,437 @@
...
+
+    @property
+    def name(self) -> str:
+        """Display name prefixing this filter's metric keys."""
+        return self._name
+
+    @property
+    def cache_key(self) -> str:
+        """Every parameter that shapes the mask, equal keys must mean equal masks."""
+        return "region:" + ",".join(sorted(self.region)) + f":{self.margin:g}"
```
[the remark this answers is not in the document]
amadeuszsz: Confirmed, the key carries the region and the margin but not the provider, so two filters with different providers would share one mask set. The fix belongs in the provider: it has to name its own maps in a config-derived way, and that class lives in #99, which is approved and sits below this PR, so it comes with the rebase rather than here.

I tried a guard local to this file first and dropped it. The only provider trait comparable through today's public API is the parsed map object, and keeping that in a module-level registry leaks between independent consumers in the same process, so it started rejecting unrelated filters. Provider objects themselves are not comparable either, every filter interpolates the node separately and gets its own instance.

Nothing is mis-sliced meanwhile: all bundled configs point every filter at one provider node.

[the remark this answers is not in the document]
amadeuszsz: The resolver now names the corpus it serves (`OsmPathResolver.cache_key`, `t4:<data_root>` for the T4 resolver), the provider passes it through, and both `RegionFilter.cache_key` and `CollisionFilter.cache_key` include it. Equal configurations still share one mask set, two data roots no longer collide, and the reported slice name is unchanged. Test covers both directions. Addressed in https://github.com/tier4/autoware-ml/pull/101/commits/e6cadf8ba781caed96d49a2ea0ac47c57767c88f.

autoware_ml/metrics/filters.py
```diff
@@ -0,0 +1,449 @@
...
+    def _region_for_pose(self, scene_token: str, x: float, y: float, heading: float):
+        """The ego collision area in the map frame for one ego pose.
+
+        The suite calls :meth:`keep` for the ground truth and the predictions of
+        the same frame in sequence and the area depends only on the pose, so one
+        cached entry serves both calls.
+        """
+        lanelet_map = self.map_provider.get(scene_token)
+        ego = ego_collision_agent(lanelet_map, x, y, heading, self.max_speed_mps, self.vehicle)
+        region = wheeled_reachable_region(ego, self.params, lanelet_map.region_union(self.region))
```
mojomex: Perf: this passes the whole scene's drivable union, and `wheeled_reachable_region` seals it with a buffer on every call. On a real 19 MB gen2 map (575k m² drivable) one ego pose costs 1.6 s, 1.35 s of which is that buffer; clipping the union to ego's reach box first (as `EgoReachability.__init__` does) brings it to 0.05 s. Once per frame at 1.6 s makes the filter the slowest part of a test run, so please clip here.

autoware_ml/metrics/filters.py
```diff
@@ -0,0 +1,449 @@
...
+            context: Per-frame values with the ego pose and scene token.
+
+        Returns:
+            Boolean mask of elements inside the collision area.
+        """
+        xyz = np.asarray(xyz, dtype=np.float64)
+        if xyz.shape[0] == 0:
+            return np.zeros((0,), dtype=bool)
+        region = self._ego_region(context)
+        if region.is_empty:
```
mojomex: Safety: when ego's pose is off the drivable region (map gap, localization offset) `available()` is still True and this returns all-False for GT and predictions alike, so the frame enters the collision slice with an empty in-path set and no FN can be scored there. Checked: ego 10 m off a road, objects on the road beside it -> `available()` True, mask `[False, False]`. Suggest `available()` builds the (cached) region and reports False when it is empty, so the exclusion is tallied like a missing map instead of silently zeroing risk.

autoware_ml/tests/metrics/test_region_filter.py
```diff
@@ -0,0 +1,444 @@
...
+        vehicle=EGO,
+        region=["road"],
+        max_speed_mps=10.0,
+        params=ReachabilityParams(horizon_s=4.0, dt_s=0.1),
+    )
+    context = {"ego2global": np.eye(4), "scene_token": "scene"}
+    xyz = np.array(
+        [
+            [10.0, 0.0, 0.0],  # ahead on the centerline, reachable -> keep
+            [-10.0, 0.0, 0.0],  # behind ego -> drop
```
mojomex: This is dropped only because the road ends at x = 0. With the road extended behind ego the same point is kept (reverse reach at v_max, checked: `[True, True, True]` for x = 10, -10, -30 on a road spanning [-50, 50]). Please make the comment say off-road, or the test passes for the wrong reason.

autoware_ml/metrics/filters.py
```diff
@@ -0,0 +1,449 @@
...
+
+    Args:
+        boxes: Box rows ``(N, 7+)`` in base_link.
+        ego2global: 4x4 ego-to-global transform of the frame.
+
+    Returns:
+        One shapely footprint polygon per box, in the map frame.
+    """
+    footprints = []
+    for box in boxes:
```
mojomex: `bev_corners_batch` already exists; a single `(N, 4, 2)` call plus one matrix transform would replace this loop and the per-row `np.stack([bev_corners(row) ...])` in `CorridorFilter.keep`.
Also, implementing `bev_corners` through `bev_corners_batch` so their impls cannot diverge would be good.

autoware_ml/tests/metrics/test_region_filter.py
mojomex: To make this test interesting, having  a point on the road far away from the ego (out of its reachable region) would be nice.

autoware_ml/metrics/geometry/lanelet.py
```diff
@@ -490,9 +490,15 @@ class OsmPathResolver(Protocol):
...
-    decide whether the scene can take part in a map-based slice.
+    decide whether the scene can take part in a map-based slice. On top of that
+    it names the corpus it resolves against, so consumers can tell two providers
+    apart without reading a map.
     """
 
+    @property
+    def cache_key(self) -> str:
+        """The set of maps this resolver serves, e.g. the dataset root."""
+
```
mojomex, line 503: Missing a return value?

autoware_ml/metrics/geometry/lanelet.py
```diff
@@ -519,6 +525,17 @@ def __init__(self, resolve_osm: OsmPathResolver) -> None:
...
+    @property
+    def cache_key(self) -> str:
+        """The set of maps this provider serves, taken from its resolver.
+
+        Consumers that group work by configuration, such as the evaluation
+        filters, have to tell two providers apart. The resolver knows which
+        corpus it points at, so the provider passes that through rather than
+        inventing an identity of its own.
+        """
+        return self._resolve_osm.cache_key
```
mojomex: Declaring the class as `@dataclass(frozen=True)` where possible gives you `hash` adn `__eq__` for free, so this extra function should not be needed.

autoware_ml/metrics/geometry/lanelet.py
```diff
@@ -565,6 +582,11 @@ def __init__(self, data_root: str) -> None:
         """
         self.data_root = str(data_root)
 
+    @property
+    def cache_key(self) -> str:
```
mojomex, line 590: Same dataclass comment as above

autoware_ml/metrics/filters.py
```diff
@@ -0,0 +1,452 @@
...
+    return footprints
+
+
+def _is_boxes(elements: np.ndarray) -> bool:
+    """Elements are detection boxes ``[cx,cy,cz,dx,dy,dz,yaw,...]`` rather than bare points."""
+    return elements.ndim == 2 and elements.shape[1] >= 7
+
+
+def ego_collision_agent(
+    lanelet_map,
```
mojomex: This would benefit from a type annotation (`LaneletMap`?)

autoware_ml/metrics/filters.py
```diff
@@ -0,0 +1,452 @@
...
+    x: float,
+    y: float,
+    heading: float,
+    max_speed_mps: float,
+    vehicle: VehicleGeometry,
+) -> Agent:
+    """Ego as a wheeled agent of the collision model, one recipe for every consumer.
```
mojomex:
```suggestion
    """Builds a wheeled agent representing the ego for use in collision
    calculations.
```

autoware_ml/metrics/filters.py
```diff
@@ -0,0 +1,452 @@
...
+
+    The collision area trim (:class:`CollisionFilter`) and the collision TTC
+    provider build ego through this single function, so the two can never drift:
+    the speed is the lanelet speed limit at ego's position with ``max_speed_mps``
+    as the off-map fallback. Callers hold the ego pose in different forms, so
+    they read it out of their transform with :func:`ego2global_pose` first. That
+    pose is base_link, the rear axle centre, which is where the vehicle
+    description measures the body from, and the description's steering limit is
+    the floor of ego's turn radius.
```
mojomex: Not sure this prose is necessary.

autoware_ml/metrics/filters.py
```diff
@@ -0,0 +1,452 @@
...
+def signed_number_token(value: float) -> str:
+    """Key-safe token that always spells the sign, e.g. ``plus0p2`` or ``minus0p2``.
+
+    Args:
+        value: Signed number to encode.
+
+    Returns:
+        The token, sign word first.
+    """
+    return ("minus" if value < 0.0 else "plus") + number_token(abs(value))
```
mojomex, line 162: (nit) extra points for assigning neither for `+/-0`

autoware_ml/tests/metrics/test_region_filter.py
```diff
@@ -0,0 +1,445 @@
...
+    front_overhang=1.0,
+    rear_overhang=1.1,
+    wheel_tread=1.64,
+    left_overhang=0.18,
+    right_overhang=0.18,
+    max_steer_angle=0.64,
+)
+
+_REAL_MAP = (
+    "data/t4dataset/db_j6gen2_v2/13cabeac-a81b-4a57-ae31-d9520a442729/0/map/lanelet2_map.osm"
```
mojomex: I'd prefer a small hand-crafted or cut out map included in this repo - as it is, CI cannot run these tests which is kinda sad.

mojomex: (also it refers to one of our internal datasets, and we want to keep references out of public scope)

autoware_ml/metrics/filters.py
```diff
@@ -0,0 +1,452 @@
...
+
+
+class RegionFilter(MetricFilter):
+    """Keep elements inside a union of lanelet2 regions.
```
mojomex:
```suggestion
    """Keep only elements inside a union of lanelet2 regions.
```

autoware_ml/metrics/filters.py
```diff
@@ -0,0 +1,452 @@
...
+        self,
+        region: list[str] | tuple[str, ...],
+        map_provider: LaneletMapProvider,
+        margin: float = 0.0,
+        name: str | None = None,
+    ) -> None:
+        """Validate the region tokens and derive the display name.
+
+        Args:
+            region: Literal lanelet2 tokens (lanelet ``subtype`` or area
```
mojomex:
```suggestion
            region: List of literal lanelet2 tokens (lanelet ``subtype`` or area
```

autoware_ml/metrics/filters.py
```diff
@@ -0,0 +1,452 @@
...
+
+    @property
+    def cache_key(self) -> str:
+        """Every parameter that shapes the mask, equal keys must mean equal masks.
+
+        The provider is part of that: a suite keeps one mask set per key, so two
+        filters that read different maps must not share it.
+        """
+        region = ",".join(sorted(self.region))
+        return f"region:{region}:{self.margin:g}:{self.map_provider.cache_key}"
```
mojomex, line 238: Prefer frozen dataclass for robustness where possible

autoware_ml/metrics/filters.py
```diff
@@ -0,0 +1,452 @@
...
+            context: Per-frame values with the scene token.
+
+        Returns:
+            Whether the scene has a lanelet map.
+        """
+        return self.map_provider.available(context["scene_token"])
+
+
+class CorridorFilter(MetricFilter):
+    """Keep elements inside a straight corridor ahead of ego.
```
mojomex:
```suggestion
    """Keep only elements inside a straight corridor ahead of ego.
```

autoware_ml/metrics/filters.py
```diff
@@ -0,0 +1,452 @@
...
+
+    @property
+    def name(self) -> str:
+        """Display name prefixing this filter's metric keys."""
+        return self._name
+
+    @property
+    def cache_key(self) -> str:
+        """Every parameter that shapes the mask, equal keys must mean equal masks."""
+        return f"corridor:w{self.width_m:g}"
```
mojomex, line 306: same as above

autoware_ml/metrics/filters.py
```diff
@@ -0,0 +1,452 @@
...
+            if far <= 0.0:
+                return np.zeros(xyz.shape[0], dtype=bool)
+            strip = shapely.box(0.0, -half_width, far, half_width)
+            footprints = shapely.polygons(corners)
+            return np.asarray(shapely.intersects(strip, footprints), dtype=bool)
+        return (xyz[:, 0] >= 0.0) & (np.abs(xyz[:, 1]) <= half_width)
+
+
+class CollisionFilter(MetricFilter):
+    """Keep elements in the ego's collision area, the filter form of the collision model.
```
mojomex:
```suggestion
    """Keep only elements in the ego's collision area, the filter form of the collision model.
```

autoware_ml/metrics/filters.py
```diff
@@ -0,0 +1,452 @@
...
+
+    The collision area is everything ego could collide with within the horizon
+    at the max map-legal speed under bounded steering, clipped to the road
+    lanelets so it follows the road on bends. A detection box is kept when its
+    footprint meets that region, a segmentation point when it lies inside it.
+    Planner-independent (never the driven path), and the same collision model
+    the criticality metrics use, so any other metric can be reported
+    in-path too.
```
mojomex, line 346: I realize that for points this would be very hard, but intuitively I'd want to assume the worst case where agents also move. This filter assumes static world except for ego.

autoware_ml/metrics/filters.py
```diff
@@ -0,0 +1,452 @@
...
+            name: Display name prefixing the metric keys.
+        """
+        if not region:
+            raise ValueError("CollisionFilter needs at least one road lanelet region token.")
+        unknown = sorted(set(str(token) for token in region) - KNOWN_REGION_TOKENS)
+        if unknown:
+            raise ValueError(
+                f"Unknown lanelet2 region tokens {unknown}, known tokens: "
+                f"{sorted(KNOWN_REGION_TOKENS)}."
+            )
```
mojomex: Same as above, might better be factored out into helper function

autoware_ml/metrics/filters.py
```diff
@@ -0,0 +1,452 @@
...
+        region = ",".join(sorted(self.region))
+        params = self.params
+        return (
+            f"collision-area:v{self.max_speed_mps:g}"
+            f":b{self.vehicle.front:g}+{self.vehicle.rear:g}x{self.vehicle.width:g}"
+            f"r{self.vehicle.min_turn_radius:g}"
+            f":h{params.horizon_s:g},{params.dt_s:g},{params.max_lateral_accel_mps2:g}"
+            f",{params.min_radius_m:g},{params.arc_samples}:{region}"
+            f":{self.map_provider.cache_key}"
+        )
```
mojomex, line 389: Same dataclass comment as above

autoware_ml/metrics/filters.py
```diff
@@ -0,0 +1,452 @@
...
+
+        Returns:
+            Boolean mask of elements inside the collision area.
+        """
+        xyz = np.asarray(xyz, dtype=np.float64)
+        if xyz.shape[0] == 0:
+            return np.zeros((0,), dtype=bool)
+        region = self._ego_region(context)
+        if region.is_empty:
+            return np.zeros(xyz.shape[0], dtype=bool)
```
mojomex, line 430: (edge case, opinion) When ego's pose is off the drivable region (map inaccurate, localization inaccurate) `available()` is still True and this returns all-False for GT and predictions alike, so the frame enters the collision slice with an empty in-path set and no FN can be scored there. Might be safer to treat map as unavailable here as the ego should never be in non-drivable areas in the first place.

autoware_ml/metrics/filters.py
amadeuszsz: Confirmed. Added `localized_surface` in the reachability module and both the filter and `EgoReachability` clip through it, so they cannot drift. On a synthetic 342k m2 surface of 1900 lanelet polygons the region build drops from 226 ms to 36 ms, and the clipped region is identical to the unclipped one (0 m2 symmetric difference at three ego poses) because that bound is the reachable set's own radius.

Addressed in https://github.com/tier4/autoware-ml/pull/101/commits/dd82c063e91b4b389d241bbeed1f46c1f077278c.

autoware_ml/tests/metrics/test_region_filter.py
amadeuszsz: You are right, it was dropped for being off the road, not for being behind ego. Comment fixed and I added the case you asked for: a point at x = 48, on the road but past 10 m/s * 4 s of reach.

Addressed in https://github.com/tier4/autoware-ml/pull/101/commits/dd82c063e91b4b389d241bbeed1f46c1f077278c.

autoware_ml/metrics/filters.py
amadeuszsz: Done. `box_footprints_map` and `CorridorFilter.keep` now take one `(N, 4, 2)` call, and `bev_corners` delegates to `bev_corners_batch`. That also removed a live contradiction: the single version documented its corner order as clockwise, the batch version documented the same order as counter-clockwise.

Addressed in https://github.com/tier4/autoware-ml/pull/101/commits/dd82c063e91b4b389d241bbeed1f46c1f077278c.

autoware_ml/metrics/geometry/lanelet.py
amadeuszsz: No, it is a `typing.Protocol`, so the members only declare the shape a resolver must have for the type checker. Nothing instantiates it (Python refuses) and nothing calls them, the real implementations are `T4LaneletMapResolver` and the test stubs. A docstring is a valid body, which is why it compiles. Added explicit `...` bodies to all three members so it no longer reads as an omission.

Addressed in https://github.com/tier4/autoware-ml/pull/101/commits/dd82c063e91b4b389d241bbeed1f46c1f077278c.

autoware_ml/metrics/geometry/lanelet.py
amadeuszsz: Made it a frozen dataclass. Kept `cache_key` though: it is a mask identity, not an object identity. It deliberately excludes the display name so two filters over the same geometry with different report names share one computed mask, and `MetricSuite` relies on that (it groups by key, then catches duplicate report keys by name). Generated `__eq__` would split those groups and compute the same mask twice. It is also consumed by `base.py`, which is already merged in #98.

Addressed in https://github.com/tier4/autoware-ml/pull/101/commits/dd82c063e91b4b389d241bbeed1f46c1f077278c.

autoware_ml/metrics/geometry/lanelet.py
amadeuszsz: Frozen dataclass here too. Same reason as above for keeping `cache_key`.

Addressed in https://github.com/tier4/autoware-ml/pull/101/commits/dd82c063e91b4b389d241bbeed1f46c1f077278c.

autoware_ml/metrics/filters.py
amadeuszsz: Done, `lanelet_map: LaneletMap`.

Addressed in https://github.com/tier4/autoware-ml/pull/101/commits/dd82c063e91b4b389d241bbeed1f46c1f077278c.

autoware_ml/metrics/filters.py
amadeuszsz: Dropped. The rear-axle and steering-limit rationale lives on `VehicleGeometry` and in the metrics doc.

Addressed in https://github.com/tier4/autoware-ml/pull/101/commits/dd82c063e91b4b389d241bbeed1f46c1f077278c.

autoware_ml/metrics/filters.py
amadeuszsz: Done, zero has no direction so it carries no sign word now.

Addressed in https://github.com/tier4/autoware-ml/pull/101/commits/dd82c063e91b4b389d241bbeed1f46c1f077278c.

autoware_ml/tests/metrics/test_region_filter.py
amadeuszsz: Agreed on both counts. The file already had a small inline OSM fixture for the speed-limit tests, so I gave it a walkway lanelet next to the road one and pointed the region-loading test at it. `_REAL_MAP` and the `skipif` are gone, so the test runs in CI and the internal path leaves this file. Note `tests/datamodule/test_frame_meta.py` still carries that dataset name in path strings, outside this PR, if you want it cleaned there too.

Addressed in https://github.com/tier4/autoware-ml/pull/101/commits/dd82c063e91b4b389d241bbeed1f46c1f077278c.

autoware_ml/metrics/filters.py
amadeuszsz: Done, frozen dataclass with the validation and name derivation in `__post_init__`. Kept `cache_key` for the reason in the thread on `LaneletMapProvider`.

Addressed in https://github.com/tier4/autoware-ml/pull/101/commits/dd82c063e91b4b389d241bbeed1f46c1f077278c.

autoware_ml/metrics/filters.py
amadeuszsz: Done, frozen dataclass.

Addressed in https://github.com/tier4/autoware-ml/pull/101/commits/dd82c063e91b4b389d241bbeed1f46c1f077278c.

autoware_ml/metrics/filters.py
amadeuszsz: Deliberate. This filter is a geometric slice, what lies in ego's reachable area now. The moving-object worst case is what the criticality metrics compute: `CollisionTTC` propagates every object's own reachable set against ego's and the metrics weight by the resulting TTC. Doing it here would duplicate that, and for segmentation points it is not definable since a bare point has no class and so no speed.

autoware_ml/metrics/filters.py
amadeuszsz: Factored the region-token validation both filters copied into `validated_region`. Reading "same as above" as that duplicate, tell me if you meant something else.

Addressed in https://github.com/tier4/autoware-ml/pull/101/commits/dd82c063e91b4b389d241bbeed1f46c1f077278c.

autoware_ml/metrics/filters.py
amadeuszsz: Done, frozen dataclass with keyword-only fields so the call shape is unchanged.

Addressed in https://github.com/tier4/autoware-ml/pull/101/commits/dd82c063e91b4b389d241bbeed1f46c1f077278c.

autoware_ml/metrics/filters.py
amadeuszsz: Good catch, that was a scoring hole: no false negative could ever be counted in a slice whose in-path set is empty. `available()` now also returns False when the collision area comes back empty, so the frame leaves the slice entirely like a map-less scene does. It reuses the cached region, so it costs nothing.

Addressed in https://github.com/tier4/autoware-ml/pull/101/commits/dd82c063e91b4b389d241bbeed1f46c1f077278c.

autoware_ml/metrics/geometry/lanelet.py
mojomex: TIL

mojomex (approved):

PR #102, opened 2026-08-25
feat(metrics): collision TTC provider and criticality metrics

amadeuszsz:
## Summary

Criticality metrics B1 and B2 on top of the reachability collision model.

- `DetectionCollisionAdapter` bridges per-frame detection boxes to the reachability engine: map-frame transforms, class to reachable-set kind, drivable polygon and per-lanelet speed limits from the scene's lanelet map, one TTC per box.
- `CriticalFPFN` (B1) reports critical false positives (phantom braking) and critical false negatives (unseen obstacles) separately as a function of the confidence threshold, class-agnostic matching, criticality is a finite TTC.
- `CollisionWeightedMeanAP` (B2) weights every object's precision and recall contribution by `e^(-decay * TTC)`, reported alongside the unweighted mAP.
- Frames without a lanelet map carry no TTC and are excluded from both metrics, with the coverage logged by the suite.

## Related Issues

Stacked on #101.

## Checklist

## Additional Notes

The detection suite invokes the adapter once per frame through the collision provider hook added in #98, both metrics share the resulting TTC.

## Additional Log and Media

`pytest autoware_ml/tests/{metrics,configs,datamodule}`: 176 passed, 1 skipped at this branch head.

### DIFF
--- a/autoware_ml/metrics/detection3d/collision.py
+++ b/autoware_ml/metrics/detection3d/collision.py
@@ -25,6 +25,12 @@ so the body splits evenly along the heading, a living agent expands a disc acros
 the widest extent the box reports, and a static agent keeps its footprint. The
 metrics turn TTC into the collision weight or the critical set.
 
+Both sides of a collision are clipped to the mapped drivable area, so a wheeled
+agent whose reference point sits outside it cannot move and is scored as
+unreachable. A vehicle in an unmapped driveway is therefore not counted, even
+with part of its body already on the road. Map coverage is what fixes that, not
+the speed fallback: off the map the speed is only the ``max_speed_mps`` default.
+
 One box costs one propagation, and a detection head emits hundreds per frame:
 200 predictions inside 60 m take about 15 s for a single frame. Metrics that
 only read confident predictions declare a ``ttc_score_floor`` so the suite can
--- a/autoware_ml/metrics/detection3d/collision.py
+++ b/autoware_ml/metrics/detection3d/collision.py
@@ -23,11 +23,16 @@ in (off-map they fall back to ``max_speed_mps``), living agents run at a per-cla
 Moving agents carry half their box width as the collision body radius, static
 agents use the full footprint. The metrics turn TTC into the collision weight
 or the critical set.
+
+One box costs one propagation, and a detection head emits hundreds per frame:
+200 predictions inside 60 m take about 15 s for a single frame. Metrics that
+only read confident predictions declare a ``ttc_score_floor`` so the suite can
+skip the rest, which is why the provider takes the scores and that floor.
 """
 
 from __future__ import annotations
 
-from math import inf
+from math import inf, nan
 from typing import Any
 
 import numpy as np
@@ -130,18 +135,31 @@ class CollisionTTC:
         return self.map_provider.available(scene_token)
 
     def per_box_ttc(
-        self, boxes: np.ndarray, labels: np.ndarray, ego2global: Any, scene_token: str
+        self,
+        boxes: np.ndarray,
+        labels: np.ndarray,
+        ego2global: Any,
+        scene_token: str,
+        scores: np.ndarray | None = None,
+        score_floor: float = 0.0,
     ) -> np.ndarray:
         """TTC (seconds, ``inf`` = unreachable) for each base_link box in the frame.
+
+        Propagation runs per box and dominates the suite's cost, so a caller that
+        knows no active metric reads the low-score tail passes ``scores`` and the
+        floor below which it can be skipped. Skipped boxes come back ``nan``.
 
         Args:
             boxes: Box rows ``[cx, cy, cz, dx, dy, dz, yaw, ...]`` in base_link.
             labels: Integer class labels aligned with ``boxes``.
             ego2global: 4x4 ego-to-global transform of the frame.
             scene_token: Scene identifier resolving to the lanelet map.
+            scores: Detection scores aligned with ``boxes``, for the floor. Ground
+                truth has none, so it is always propagated.
+            score_floor: Lowest score worth propagating, ``0.0`` propagates every box.
 
         Returns:
-            Per-box TTC array of shape ``(N,)``.
+            Per-box TTC array of shape ``(N,)``, ``nan`` where the floor skipped a box.
         """
         boxes = np.asarray(boxes, dtype=np.float64)
         labels = np.asarray(labels).astype(int)
@@ -160,30 +178,64 @@ class CollisionTTC:
         centroids = np.array([[p.centroid.x, p.centroid.y] for p in footprints], dtype=np.float64)
 
         for index in range(boxes.shape[0]):
-            name = self.class_names[int(labels[index])]
-            kind = self.kinds[name]
-            cx, cy = centroids[index, 0], centroids[index, 1]
-            length, width = float(boxes[index, 3]), float(boxes[index, 4])
-            if kind == AgentKind.STATIC:
-                agent = Agent.static(cx, cy, footprints[index])
-            elif kind == AgentKind.WHEELED:
-                # Max speed = the speed limit of the lanelet the agent is in. A box
-                # is centred, so its body splits evenly around the reference point,
-                # and it carries no steering geometry, so the friction floor stands in.
-                speed = lanelet_map.speed_at(cx, cy, self.max_speed_mps)
-                agent = Agent.wheeled(
-                    cx,
-                    cy,
-                    float(boxes[index, 6]) + ego.heading,
-                    speed,
-                    front=length / 2.0,
-                    rear=length / 2.0,
-                    width=width,
-                    min_turn_radius=self.params.min_radius_m,
-                )
-            else:
-                # A living agent runs at its class speed in any direction, so its
-                # body is a disc across the widest extent the box reports.
-                agent = Agent.living(cx, cy, self.living_speeds[name], max(length, width) / 2.0)
+            if scores is not None and scores[index] < score_floor:
+                # Never read by any active metric: NaN so a metric that reads it
+                # anyway fails loud instead of taking it for "unreachable".
+                ttc[index] = nan
+                continue
+            agent = self._agent_for_box(
+                boxes[index],
+                int(labels[index]),
+                footprints[index],
+                centroids[index],
+                lanelet_map,
+                ego.heading,
+            )
             ttc[index] = frame.time_to_collision(agent)
         return ttc
+
+    def _agent_for_box(
+        self,
+        box: np.ndarray,
+        label: int,
+        footprint: Any,
+        centroid: np.ndarray,
+        lanelet_map: Any,
+        ego_heading: float,
+    ) -> Agent:
+        """The collision agent one detection box stands for, by its class kind.
+
+        Args:
+            box: Box row ``[cx, cy, cz, dx, dy, dz, yaw, ...]`` in base_link.
+            label: Integer class label of the box.
+            footprint: The box's map-frame footprint polygon.
+            centroid: The footprint's map-frame ``(x, y)`` centroid.
+            lanelet_map: The scene's parsed lanelet map, for the local speed limit.
+            ego_heading: Ego map-frame heading, added to the box's own yaw.
+
+        Returns:
+            The agent for this box.
+        """
+        name = self.class_names[label]
+        kind = self.kinds[name]
+        cx, cy = float(centroid[0]), float(centroid[1])
+        length, width = float(box[3]), float(box[4])
+        if kind == AgentKind.STATIC:
+            return Agent.static(cx, cy, footprint)
+        if kind == AgentKind.WHEELED:
+            # Max speed = the speed limit of the lanelet the agent is in. A box is
+            # centred, so its body splits evenly around the reference point, and it
+            # carries no steering geometry, so the friction floor stands in.
+            return Agent.wheeled(
+                cx,
+                cy,
+                float(box[6]) + ego_heading,
+                lanelet_map.speed_at(cx, cy, self.max_speed_mps),
+                front=length / 2.0,
+                rear=length / 2.0,
+                width=width,
+                min_turn_radius=self.params.min_radius_m,
+            )
+        # A living agent runs at its class speed in any direction, so its body is a
+        # disc across the widest extent the box reports.
+        return Agent.living(cx, cy, self.living_speeds[name], max(length, width) / 2.0)
--- a/autoware_ml/metrics/detection3d/collision.py
+++ b/autoware_ml/metrics/detection3d/collision.py
@@ -67,7 +67,9 @@ DEFAULT_KINDS: dict[str, AgentKind] = {
     "traffic_cone": AgentKind.STATIC,
     "debris": AgentKind.STATIC,
     "bicycle_rack": AgentKind.STATIC,
-    "vehicle_extension": AgentKind.STATIC,
+    # A trailer or towed body moves with whatever pulls it, so it gets the wheeled
+    # worst case rather than staying where the frame found it.
+    "vehicle_extension": AgentKind.WHEELED,
 }
 # Living "reasonable run" speeds (m/s). Wheeled speed comes from the lanelet map.
 DEFAULT_LIVING_SPEEDS: dict[str, float] = {"pedestrian": 3.0, "animal": 4.0, "bicycle": 6.0}
--- a/autoware_ml/metrics/detection3d/collision.py
+++ b/autoware_ml/metrics/detection3d/collision.py
@@ -20,9 +20,10 @@ assigns every class its reachable-set kind, pulls the drivable polygon and the
 per-lanelet speed limits from the scene's lanelet map, and returns one TTC per
 box. Ego and wheeled agents move at the ``speed_limit`` of the lanelet they are
 in (off-map they fall back to ``max_speed_mps``), living agents run at a per-class speed.
-Moving agents carry half their box width as the collision body radius, static
-agents use the full footprint. The metrics turn TTC into the collision weight
-or the critical set.
+A wheeled box sweeps its own rectangle, its reference point being the box centre
+so the body splits evenly along the heading, a living agent expands a disc across
+the widest extent the box reports, and a static agent keeps its footprint. The
+metrics turn TTC into the collision weight or the critical set.
 
 One box costs one propagation, and a detection head emits hundreds per frame:
 200 predictions inside 60 m take about 15 s for a single frame. Metrics that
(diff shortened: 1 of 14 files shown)


autoware_ml/metrics/geometry/reachability.py
```diff
@@ -0,0 +1,386 @@
...
+            raise ValueError("ego reachability needs a drivable polygon to clip against.")
+        self.ego = ego
+        self.params = params
+        # Floored with a tolerance: an exact multiple of dt_s keeps its final step and a
+        # non-divisible horizon never gains a step beyond it.
+        self.steps = int(params.horizon_s / params.dt_s + 1e-9)
+        reach = ego.speed * params.horizon_s + ego.body_radius
+        self._drivable = drivable.intersection(
+            box(ego.x - reach, ego.y - reach, ego.x + reach, ego.y + reach)
+        )
```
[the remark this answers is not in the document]
amadeuszsz: Confirmed, the arc feasibility check made ego's reach clip destructive for incoming objects. Fixed in #100: ego fronts keep the ego-local clip, while each wheeled object's arcs are checked against the full drivable surface localized to that object's own reach box, prepared once per object. Regression test with an oncoming vehicle at 50 m against a 41 m ego clip asserts a finite TTC around 2.5 s.

autoware_ml/metrics/detection3d/suite.py
```diff
@@ -79,8 +201,29 @@ def _warn_on_range_class_caps(self) -> None:
...
+        The collision TTC needs the ego pose and scene token exactly like the
+        region filters do. Declaring them here makes a missing key fail loud at
+        the first batch instead of silently reporting zero coverage.
+
+        Returns:
+            The required eval-output key names.
+        """
+        keys = super().required_keys()
+        if self.collision is not None:
+            keys = tuple(dict.fromkeys(keys + ("ego2global", "scene_token")))
```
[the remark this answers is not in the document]
amadeuszsz: This is the stack split: the detection eval output forwards ego2global and scene_token, and the detection collation maps declare them, in #107 (`detection_eval_output` and the task base configs there). Nothing in this PR runs before that lands.

autoware_ml/configs/tasks/detection3d/bevfusion/camera_lidar_voxel0075_second_secfpn_54m_nuscenes.yaml
```diff
@@ -73,6 +73,9 @@ datamodule:
         point_cloud_range: ${point_cloud_range}
       - _target_: autoware_ml.transforms.boxes3d.filters.ObjectRangeFilter
         point_cloud_range: ${point_cloud_range}
+      - _target_: autoware_ml.transforms.boxes3d.filters.ObjectRangeMinPointsFilter
+        range_radius: [0.0, 60.0]
+        min_num_points: 1
```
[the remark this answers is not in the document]
amadeuszsz: Right, the filters bound by radius while the training range is a square, so the corners past 130 m (173.8 m at the T4 corner, 76 m on nuScenes) were never point checked. The far filters now have no outer bound (`range_radius: [60.0, .inf]`, the single nuScenes filters `[0.0, .inf]`), so every box inside the rectangle is checked, and the same filters now run on the validation pipelines too. The filter wiring lives in the multisweep PR, so that is where it landed: https://github.com/tier4/autoware-ml/pull/112/commits/4d7c105a2b820f173469819e3f2c337a439f682d.

autoware_ml/metrics/detection3d/collision.py
```diff
@@ -0,0 +1,187 @@
...
+        centroids = np.array([[p.centroid.x, p.centroid.y] for p in footprints], dtype=np.float64)
+
+        for index in range(boxes.shape[0]):
+            name = self.class_names[int(labels[index])]
+            kind = self.kinds[name]
+            cx, cy = centroids[index, 0], centroids[index, 1]
+            length, width = float(boxes[index, 3]), float(boxes[index, 4])
+            if kind == AgentKind.STATIC:
+                agent = Agent.static(cx, cy, footprints[index])
+            elif kind == AgentKind.WHEELED:
```
mojomex: A wheeled GT whose centre lies off the drivable polygon gets an empty reachable set and TTC = inf, i.e. weight 0 and never critical, regardless of distance. Checked with the adapter on a 7 m road strip: a car 8 m ahead with its centre 1 m outside the edge, nose toward the road, scores inf; the same car 1.5 m further in scores 0.2 s. Vehicles pulling out of driveways / parking lots, and lanelet polygons that undershoot the paved width, both land in that bucket. Suggest a fallback for off-surface wheeled agents (e.g. score them as living at the lane speed, or at least their stopped body) rather than inf.

autoware_ml/metrics/detection3d/collision.py
```diff
@@ -0,0 +1,187 @@
...
+        drivable = lanelet_map.region_union(self.region)
+        ego_x, ego_y, ego_heading = ego2global_pose(ego2global)
+        ego = ego_collision_agent(
+            lanelet_map, ego_x, ego_y, ego_heading, self.max_speed_mps, self.vehicle
+        )
+        frame = EgoReachability(ego, drivable, self.params)
+        footprints = box_footprints_map(boxes, ego2global)
+        centroids = np.array([[p.centroid.x, p.centroid.y] for p in footprints], dtype=np.float64)
+
+        for index in range(boxes.shape[0]):
```
mojomex, line 189: This runs the reachability engine on every prediction, including the hundreds of low-score boxes a head emits. Measured: 200 random predictions within 60 m take 8.2 s for one frame (ego steps + 59 finite objects). CriticalFPFN only reads scores >= its lowest confidence; consider a score floor for pred TTC, or at least document the cost. Nit: the per-kind agent construction would read better as a small `_agent_for_box` helper.

autoware_ml/metrics/detection3d/collision.py
```diff
@@ -0,0 +1,187 @@
...
+
+"""Detection-box to reachability time-to-collision adapter.
+
+Bridges the per-frame detection boxes (base_link ``[cx, cy, cz, dx, dy, dz, yaw,
+...]``) to the reachability engine: transforms ego and each box to the map frame,
+assigns every class its reachable-set kind, pulls the drivable polygon and the
+per-lanelet speed limits from the scene's lanelet map, and returns one TTC per
+box. Ego and wheeled agents move at the ``speed_limit`` of the lanelet they are
+in (off-map they fall back to ``max_speed_mps``), living agents run at a per-class speed.
+Moving agents carry half their box width as the collision body radius, static
```
mojomex: Stale: wheeled agents now sweep their box (front/rear = length/2, width), living agents use max(length, width)/2. Please update.

autoware_ml/metrics/detection3d/collision.py
```diff
@@ -0,0 +1,187 @@
...
+    "animal": AgentKind.LIVING,
+    "bicycle": AgentKind.LIVING,
+    "barrier": AgentKind.STATIC,
+    "traffic_cone": AgentKind.STATIC,
+    "debris": AgentKind.STATIC,
+    "bicycle_rack": AgentKind.STATIC,
+    "vehicle_extension": AgentKind.STATIC,
+}
+# Living "reasonable run" speeds (m/s). Wheeled speed comes from the lanelet map.
+DEFAULT_LIVING_SPEEDS: dict[str, float] = {"pedestrian": 3.0, "animal": 4.0, "bicycle": 6.0}
```
mojomex: Future work (not blocking): 6 m/s for bicycle is below what road cyclists routinely ride (8-10 m/s) and e-bike assist caps at 6.9 m/s. For a worst-case model 8 m/s seems the safer default.

autoware_ml/metrics/detection3d/collision_weighted_map.py
```diff
@@ -0,0 +1,155 @@
+"""Collision-risk-weighted mAP.
+
+Every object's contribution to precision and recall is weighted by ``w =
+e^(-decay * TTC)``, with TTC the reachability time-to-collision to ego computed
+once per frame by the suite's collision provider: each agent and ego are
+propagated at their class max speed, and TTC is the first time their reachable
+sets overlap. An object that cannot be reached within the horizon has TTC = inf
+and weight 0, so a same-speed lead or off-road scenery buys no score,
```
mojomex: Stale since the forward/reverse rework in #100: a same-speed lead now has a finite TTC and a non-zero weight. Please update.

autoware_ml/metrics/detection3d/collision.py
```diff
@@ -0,0 +1,189 @@
...
+    "train": AgentKind.WHEELED,
+    "motorcycle": AgentKind.WHEELED,
+    "pedestrian": AgentKind.LIVING,
+    "animal": AgentKind.LIVING,
+    "bicycle": AgentKind.LIVING,
+    "barrier": AgentKind.STATIC,
+    "traffic_cone": AgentKind.STATIC,
+    "debris": AgentKind.STATIC,
+    "bicycle_rack": AgentKind.STATIC,
+    "vehicle_extension": AgentKind.STATIC,
```
mojomex: `vehicle_extension` can move with the vehicle so feels like it should be `WHEELED`.

autoware_ml/metrics/detection3d/collision.py
```diff
@@ -0,0 +1,189 @@
...
+            kind = self.kinds[name]
+            cx, cy = centroids[index, 0], centroids[index, 1]
+            length, width = float(boxes[index, 3]), float(boxes[index, 4])
+            if kind == AgentKind.STATIC:
+                agent = Agent.static(cx, cy, footprints[index])
+            elif kind == AgentKind.WHEELED:
+                # Max speed = the speed limit of the lanelet the agent is in. A box
+                # is centred, so its body splits evenly around the reference point,
+                # and it carries no steering geometry, so the friction floor stands in.
+                speed = lanelet_map.speed_at(cx, cy, self.max_speed_mps)
```
mojomex: Question: our lanelet probably don't include driveways. So a car pulling out of one would default to `max_speed_mps`, and not be ignored, right? (IMO `max_speed_mps` is acceptable, being ignored not)

autoware_ml/metrics/detection3d/critical_fp_fn.py
```diff
@@ -0,0 +1,125 @@
+"""Critical FP / FN over the ego reachable set.
+
+Two numbers, never averaged: false positives in ego's path cause phantom braking
```
mojomex: Nit: `Two numbers, never averaged:` very Claude-y, please rephrase.

autoware_ml/tests/metrics/test_criticality_metrics.py
```diff
@@ -0,0 +1,165 @@
...
+    assert out["critical_fp_conf0p5"] == pytest.approx(1.0)
+
+    weighted = CollisionWeightedMeanAP(thresholds=(2.0,), decay=0.5).evaluate(state, EvalStage.TEST)
+    # The uncovered frame's phantom prediction contributes nothing: the covered
+    # frame's perfect match still scores, so cw_mAP stays finite and positive.
+    assert weighted["cw_mAP"] > 0.0
+
+
+def test_critical_fp_fn_no_coverage_is_nan() -> None:
+    import math
```
mojomex: Nit: favor top-level imports

autoware_ml/tests/metrics/test_criticality_metrics.py
```diff
@@ -0,0 +1,165 @@
...
+        class_names=("car",),
+    )
+    out = CollisionWeightedMeanAP(thresholds=(2.0,), decay=0.5).evaluate(state, EvalStage.TEST)
+    assert out["cw_mAP"] == pytest.approx(1.0)
+    assert out["cw_mAP_car"] == pytest.approx(1.0)
+
+
+def test_collision_weighted_map_unreachable_gt_has_no_weight() -> None:
+    # All GT unreachable (TTC inf) -> total GT weight 0 -> AP is NaN (nothing to score).
+    import math
```
mojomex: same as above

autoware_ml/tests/metrics/test_suite_collision.py
```diff
@@ -0,0 +1,165 @@
...
+
+    def get(self, scene_token):
+        return self._map
+
+    def available(self, scene_token):
+        return True
+
+
+def _b(cx, cy, yaw=0.0):
+    return [cx, cy, 0.0, 4.0, 2.0, 1.5, yaw, 0.0, 0.0]
```
mojomex: These are duplicated across tests, consider making them fixtures instead.

autoware_ml/tests/metrics/test_suite_collision.py
```diff
@@ -0,0 +1,165 @@
...
+            vehicle=EGO,
+            params=ReachabilityParams(horizon_s=4.0, dt_s=0.1),
+        ),
+    )
+    keys = suite.required_keys()
+    assert "ego2global" in keys and "scene_token" in keys
+
+
+def test_min_num_points_without_counts_raises() -> None:
+    import pytest
```
mojomex: same as above

mojomex (commented):
Mostly good to go, only some small feedback and one performance concern.

autoware_ml/metrics/detection3d/collision.py
amadeuszsz: Confirmed, and worse on my setup: 200 predictions inside 60 m take 15.5 s in one frame, linear in box count.

A plain score floor would be wrong for `CollisionWeightedMeanAP`, which weights every prediction and integrates the whole score range. So the floor is now a component contract: `ttc_score_floor` next to `needs_ttc`, default 0.0 so a metric that ignores it stays correct. `CriticalFPFN` declares `min(confidences)`, the suite takes the minimum over the active metrics and passes it with the scores. Skipped boxes come back `nan`, and `CriticalFPFN` raises if a confidence below its own floor ever reads one, so a misconfiguration cannot pass silently.

With `CriticalFPFN` alone at 0.3, a 120-box frame skips 33 boxes and update drops from 28.0 s to 20.6 s, with identical counts. The shipped config also wires `CollisionWeightedMeanAP`, so there the floor is 0 and nothing is skipped; the measured cost is now in the module docstring. Also split the per-kind construction into `_agent_for_box`.

Addressed in https://github.com/tier4/autoware-ml/pull/102/commits/90fd545d4ccb59a03f9c8801481dfc0c9672d651.

autoware_ml/metrics/detection3d/collision.py
amadeuszsz: Updated. A wheeled box now sweeps its own rectangle around the box centre, a living agent expands a disc across the widest extent the box reports, and a static agent keeps its footprint.

Addressed in https://github.com/tier4/autoware-ml/pull/102/commits/977537e9239195415b3a208bdd3fb6aba891ec4c.

autoware_ml/metrics/detection3d/collision_weighted_map.py
amadeuszsz: Fixed. The zero-weight example is now scenery off the drivable area or an object parked beyond the horizon, and the text says a matched-speed lead keeps a small TTC and a high weight because it may brake.

Addressed in https://github.com/tier4/autoware-ml/pull/102/commits/fbb98cb.

autoware_ml/metrics/detection3d/collision.py
amadeuszsz: Agreed, a trailer moves with whatever pulls it. Moved to WHEELED and dropped it from the static list in the reachability module docstring. Worth knowing this changes numbers: a parked trailer now gets a full reachable set, so it can turn into a critical false positive. The class table in #109 still lists it under static, I will fix that there.

Addressed in https://github.com/tier4/autoware-ml/pull/102/commits/90d1602.

autoware_ml/metrics/detection3d/collision.py
amadeuszsz: Half right, and the other half is worse than you feared. The speed does fall back to `max_speed_mps`. But the object is also ignored, and I measured it: with a 12 m mapped road, a car three metres into an unmapped driveway with its nose pointing at the road gets TTC = inf, and so does one whose body already overlaps the road by half a metre. The reason is that a wheeled agent's travel is clipped to the mapped surface and every arc has to start there, so an agent whose reference point is off the map cannot move at all.

Ego's own region is clipped the same way, so ego can never be in the driveway either, which means no clipping-based tweak catches this: only the object being allowed to travel off-map would, and that lets cars drive through buildings. So I wrote the limitation down rather than papering over it, and map coverage is the real fix. If you want a middle ground, dilating an object's drivable surface by its own body length would catch a car within one length of the road; say the word and I will add it.

Addressed in https://github.com/tier4/autoware-ml/pull/102/commits/0e65d722a971a35bb811e2d1da439998b9dea878.

autoware_ml/metrics/detection3d/critical_fp_fn.py
amadeuszsz: Rephrased: the two counts are reported separately because they mean different things, a false positive in ego's path causes phantom braking and a false negative means driving toward something nobody saw.

Addressed in https://github.com/tier4/autoware-ml/pull/102/commits/5054a40a3a6b78310fd2eee5a5041e5910fa2f64.

autoware_ml/tests/metrics/test_criticality_metrics.py
amadeuszsz: Moved to the top of the file, both here and in the sibling test.

Addressed in https://github.com/tier4/autoware-ml/pull/102/commits/ac2c40e7c03c1601e89456e9624d51b2ab7d79eb.

autoware_ml/tests/metrics/test_criticality_metrics.py
amadeuszsz: Same, moved up.

Addressed in https://github.com/tier4/autoware-ml/pull/102/commits/ac2c40e7c03c1601e89456e9624d51b2ab7d79eb.

autoware_ml/tests/metrics/test_suite_collision.py
amadeuszsz: Done. The vehicle, the class list, the box helper and the map stubs now live in `autoware_ml/tests/metrics/conftest.py` and the three collision test files import them, which removed about 150 duplicated lines.

Addressed in https://github.com/tier4/autoware-ml/pull/102/commits/894365f6b5d5c5f1f214fadc99922626b089f199.

autoware_ml/tests/metrics/test_suite_collision.py
amadeuszsz: That one was the function-level `import pytest`, now at the top of the file.

Addressed in https://github.com/tier4/autoware-ml/pull/102/commits/90fd545d4ccb59a03f9c8801481dfc0c9672d651.

mojomex (approved):

PR #103, opened 2026-08-25
feat(metrics): segmentation confusion suite over filter buckets

amadeuszsz:
## Summary

Replaces the range-only segmentation suite with `Segmentation3DConfusionMatrixMetricSuite` over (filter, range) buckets.

- One fixed-size confusion matrix per (filter, range) bucket in a single stacked `(F+1, R+1, C, C)` sum state, so memory never grows with the dataset and DDP reduction stays exact.
- Consumes per-frame `seg_frames` entries (points plus ego pose and scene token), resolving per-point filter membership at update, with map coverage counters for scenes without a lanelet map.
- Adds the raw `ConfusionMatrix` component and per-class-group folding, replaces the flat point-tensor input path, and updates the distributed and segmentation tests to the new contract.

## Related Issues

Stacked on #102.

## Checklist

- [x] You've checked our [contribution overview](https://tier4.github.io/autoware-ml/contributing/contribution-overview/).
- [x] Your PR follows our [pull request guidelines](https://tier4.github.io/autoware-ml/contributing/pull-request-guidelines/).

## Additional Notes

Models emit `seg_frames` in a follow-up PR of this stack, until then the suite is exercised by its tests only.

## Additional Log and Media

`pytest autoware_ml/tests/{metrics,configs,datamodule}`: 177 passed, 1 skipped at this branch head.

PR #104, opened 2026-08-25
feat(metrics): test-only point-cloud suite and seg_frames contract

amadeuszsz:
## Summary

Test-only point-cloud segmentation suite and the shared `seg_frames` builder.

- `Segmentation3DPointCloudMetricSuite` caches per-frame points, labels, reduced softmax scalars and raw boxes on CPU for the point-level metrics that a bounded confusion matrix cannot express, with the same filter and range axes and map coverage rules as the confusion suite.
- One accumulation serves the trained taxonomy and the behaviour class groups: labels fold at state build, the grouped uncertainty scalars reduce at update, `grouped_components` read the `grouped/` view.
- `segmentation_frames_eval_output` builds the per-frame `seg_frames` entries from batched tensors, carrying ego pose, scene token, timestamp and per-frame detection ground truth when present.

## Related Issues

Stacked on #103.

## Checklist

## Additional Notes

Runs on the test stage only, validation keeps the lightweight confusion suite. The point-level metric components consuming this state arrive next in the stack.

## Additional Log and Media

`pytest autoware_ml/tests/{metrics,configs,datamodule}`: 179 passed, 1 skipped at this branch head.


amadeuszsz:
Regarding the eval-output finding: that is the stacking order. The datamodules and `detection_eval_output` gain `ego2global` and `scene_token` forwarding in a dedicated follow-up PR of this stack, and no shipped configuration enables a map-backed filter or TTC component until the config migration PR that comes after that wiring. At this branch head the filtered detection path is not reachable from any built-in configuration.

autoware_ml/metrics/detection3d/suite.py
```diff
@@ -79,8 +201,29 @@ def _warn_on_range_class_caps(self) -> None:
...
+        The collision TTC needs the ego pose and scene token exactly like the
+        region filters do. Declaring them here makes a missing key fail loud at
+        the first batch instead of silently reporting zero coverage.
+
+        Returns:
+            The required eval-output key names.
+        """
+        keys = super().required_keys()
+        if self.collision is not None:
+            keys = tuple(dict.fromkeys(keys + ("ego2global", "scene_token")))
```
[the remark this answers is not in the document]
amadeuszsz: This is the stack split rather than a gap: the detection eval output forwards ego2global and scene_token, and the bundled detection collation maps declare them, in #107 right above this PR. Provided in https://github.com/tier4/autoware-ml/pull/107/commits/7d59d34d5bede73f55b1fb1a2e2eaf26fe839446.

autoware_ml/metrics/segmentation3d/point_cloud.py
```diff
@@ -0,0 +1,591 @@
...
+
+class Segmentation3DPointCloudMetricSuite(MetricSuite[PointCloudSegState]):
+    """Caches raw per-point segmentation data at test and exposes it per range.
+
+    Point-level metrics (error clusters, neighbourhood-tolerant error, partial
+    detection, calibration, entropy) are injected as stateless components.
+    """
+
+    prefix = "seg3d_pt"
+    _required_keys = ("seg_frames",)
```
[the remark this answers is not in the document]
amadeuszsz: Same split: FRNet and the PTv3 seg head switch to segmentation_frames_eval_output in #106, so the suite has a producer as soon as that merges. Provided in https://github.com/tier4/autoware-ml/pull/106/commits/4ea6cd24da9aa87dca5a4bb9709ccceaa5a32a42.

autoware_ml/metrics/segmentation3d/point_cloud.py
```diff
@@ -0,0 +1,584 @@
+"""Test-only point-cache segmentation suite for spatial metrics.
+
+The confusion-matrix suite is a bounded sufficient statistic for IoU / accuracy /
+precision-recall, but the driving-aware segmentation metrics need per-point
+spatial information the matrix has thrown away: spatial clusters (drivable-area
+phantoms), neighbourhood-tolerant errors, per-object grouping (partial detection)
+and the full softmax distribution (calibration, entropy).
+
+At test we therefore cache per-point data as list-states: coordinates, predicted and target
```
mojomex: This paragraph and the `update` docstring say the same thing three times (what is cached, why the softmax is reduced, how the grouped view is served). Please cut each to the facts a reader needs: what is cached per point (coord, pred, target, confidence, entropy; int16 labels), that it is test-only / single-process, and that the grouped view folds labels at state build. The design rationale is fine once.

autoware_ml/metrics/segmentation3d/point_cloud.py
```diff
@@ -0,0 +1,584 @@
...
+                torch.zeros((0, 7), dtype=torch.float32)
+                if boxes is None
+                else boxes.detach().to(dtype=torch.float32).cpu()
+            )
+            self.gt_box_labels.append(
+                torch.zeros((0,), dtype=torch.long)
+                if labels is None
+                else labels.detach().to(dtype=torch.long).cpu()
+            )
+            if self._component_filters and active:
```
mojomex, line 342: `update` is ~110 lines. Pulling this filter-mask block into a `_filter_masks(frame, coord_np, boxes_np, active) -> (point_masks, box_masks)` helper would leave `update` as validate / reduce / cache, which is what a reader expects.

autoware_ml/metrics/segmentation3d/point_cloud.py
```diff
@@ -0,0 +1,584 @@
...
+        if boxes_np.shape[0]:
+            box_keep = (
+                np.ones(boxes_np.shape[0], dtype=bool)
+                if box_region_mask is None
+                else box_region_mask
+            )
+            if metric_range is not None:
+                box_keep = box_keep & self._range_mask(boxes_np[:, :2], metric_range)
+            boxes_np, labels_np = boxes_np[box_keep], labels_np[box_keep]
+        if keep.all():
```
mojomex: The two return paths duplicate the fold and the tuple assembly. Slicing once (`arr if keep.all() else arr[keep]` for the five per-point arrays), folding once, then freezing the shared case would halve this function. I checked that the freeze does its job: a component that writes into a whole-scene frame gets a ValueError, and a second `compute()` returns identical numbers.

mojomex:
PR commit 64498e5 is buried under seemingly unrelated 3fee951, 3d0d10f, please check.

autoware_ml/metrics/segmentation3d/point_cloud.py
```diff
@@ -0,0 +1,584 @@
...
+
+
+def _reduce_scores(scores: torch.Tensor, pred: torch.Tensor) -> tuple[torch.Tensor, torch.Tensor]:
+    """Per-point (confidence, normalized entropy) from a softmax, on its device.
+
+    Confidence is the probability of the reported prediction and entropy is
+    normalized by ``log`` of the distribution's width.
+    """
+    probs = scores.clamp(1e-12, 1.0)
+    entropy = -(probs * probs.log()).sum(dim=1) / float(np.log(scores.shape[1]))
```
mojomex: Please add docstring entries for the two params.
Are `scores` already softmaxed, so guaranteed to be in `[0, 1]`?

autoware_ml/metrics/segmentation3d/point_cloud.py
```diff
@@ -0,0 +1,584 @@
...
+    """Caches raw per-point segmentation data at test and exposes it per range.
+
+    Point-level metrics (error clusters, neighbourhood-tolerant error, partial
+    detection, calibration, entropy) are injected as stateless components.
+    """
+
+    prefix = "seg3d_pt"
+    _required_keys = ("seg_frames",)
+
+    _FRAME_STATES = (
```
mojomex: (nit) `STATES` sounds weird, `KEYS` or `FIELDS` might be better.

mojomex (dismissed):
Mostly LGTM, but please address the few feedback items before merging 🙏

autoware_ml/metrics/segmentation3d/point_cloud.py
amadeuszsz: They are softmax probabilities, the head softmaxes before building `seg_frames`, but nothing said that and nothing checked it. Bad, because confidence reads the raw scores and the grouped view sums columns, so logits would pass through and give garbage. Documented it now and added a check in update next to the column one. Params documented too.

Addressed in https://github.com/tier4/autoware-ml/pull/104/commits/f6c1373a564ece0c1c909e087db4b892d809f13f

autoware_ml/metrics/segmentation3d/point_cloud.py
amadeuszsz: Renamed to `_FRAME_FIELDS`. They are the `FramePoints` fields anyway, so that reads better.

Addressed in https://github.com/tier4/autoware-ml/pull/104/commits/36f05f6164b3ba15caaa7d16b088e95438f082c8

autoware_ml/metrics/segmentation3d/point_cloud.py
amadeuszsz: Cut both. Module docstring says what is cached and that it is test only single process, the grouped view gets one short paragraph, and update keeps one line about the softmax reduction. Rationale is there once now.

Addressed in https://github.com/tier4/autoware-ml/pull/104/commits/30191c251b01bd57276aff2300389c79373fd96d

autoware_ml/metrics/segmentation3d/point_cloud.py
amadeuszsz: Pulled it into `_filter_masks`, with the column stacking in `_stack_masks` next to it. `update` is 89 lines now and reads validate, reduce, cache.

Addressed in https://github.com/tier4/autoware-ml/pull/104/commits/833c9b2ace0d9e2918900f3946d449b2c22a88e6

autoware_ml/metrics/segmentation3d/point_cloud.py
amadeuszsz: Done, one path now. The five per point arrays are sliced in one comprehension, the fold runs once, and the freeze only happens in the shared case. 28 lines shorter. Thanks for checking the freeze, that behaviour is unchanged.

Addressed in https://github.com/tier4/autoware-ml/pull/104/commits/9bcc3d664ab12f0f0b2f3da7dbedf6e53ec74d3c

PR #105, opened 2026-08-25
feat(metrics): point-level segmentation metrics

amadeuszsz:
## Summary

Point-level segmentation metric components over the point-cloud suite state (#104).

- Error structure: `ErrorClusters` (spatially clustered wrong points), `NeighbourhoodTolerantErrorRate` (errors with no correct-class neighbour within a radius) and `ConfidentErrorRate`, with the shared neighbourhood math in `spatial.py`.
- Uncertainty: `CalibrationError` and `UncertaintyUsefulness` from the reduced per-point confidence and normalized entropy.
- `PartialDetectionScore` cross-checks segmentation against the per-frame detection ground truth carried by the suite.
- Components report per trained class and per behaviour class group, on the whole scene and on every configured filter slice.

## Related Issues

Stacked on #104.

## Checklist

## Additional Notes

Purely additive components plus their suite integration tests, nothing configures them yet.

## Additional Log and Media

`pytest autoware_ml/tests/{metrics,configs,datamodule}`: 208 passed, 1 skipped at this branch head.


autoware_ml/metrics/segmentation3d/point_cloud.py
```diff
@@ -0,0 +1,482 @@
...
+
+class Segmentation3DPointCloudMetricSuite(MetricSuite[PointCloudSegState]):
+    """Caches raw per-point segmentation data at test and exposes it per range.
+
+    Point-level metrics (error clusters, neighbourhood-tolerant error, partial
+    detection, calibration, entropy) are injected as stateless components.
+    """
+
+    prefix = "seg3d_pt"
+    _required_keys = ("seg_frames",)
```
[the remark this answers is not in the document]
amadeuszsz: That is the stacking order: the models switch their build_eval_output to the seg_frames builder in the very next PR of this stack, and the configs adopt this suite in the config migration PR after that. Until then the shipped configurations keep the confusion suite, this one is exercised by its tests.

autoware_ml/metrics/segmentation3d/point_cloud.py
```diff
@@ -0,0 +1,482 @@
...
+            labels = frame.get("gt_box_labels")
+            self.gt_boxes.append(
+                torch.zeros((0, 7), dtype=torch.float32)
+                if boxes is None
+                else boxes.detach().to(dtype=torch.float32).cpu()
+            )
+            self.gt_box_labels.append(
+                torch.zeros((0,), dtype=torch.long)
+                if labels is None
+                else labels.detach().to(dtype=torch.long).cpu()
```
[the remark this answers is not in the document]
amadeuszsz: Valid, that was a silent fallback. Components now declare needs_boxes (PartialDetectionScore sets it) and the suite raises at update when a box-reading component is configured and a frame omits gt_boxes or gt_box_labels, empty tensors remain the explicit no-objects form. Covered by a new test.

autoware_ml/metrics/segmentation3d/partial_detection.py
```diff
@@ -0,0 +1,185 @@
...
+                if candidates.shape[0]:
+                    inside = candidates[_points_in_box(coord[candidates], box)]
+                else:
+                    inside = candidates
+                n = int(inside.shape[0])
+                if n < self.min_points:
+                    skipped += 1
+                    continue
+                # The detection class and its segmentation class share the index.
+                correct = int(np.sum(pred[inside] == label))
```
mojomex, line 173: The credit never reads `target`: `n` is every point inside the GT box and `correct` is every point predicted as the box class, so ground returns in the footprint (the box reaches the ground) both dilute and inflate the score.

Measured with a 1.8 m pedestrian box, 20 pedestrian points and 15 ground points inside it: all pedestrian points correct -> `pd_score_ped = 0.980` instead of 1.0; zero pedestrian points found but the ground mispredicted as pedestrian -> `0.964` instead of 0.

Please restrict both `n` and `correct` to interior points whose `target == label` (the seg GT is right there in the frame).

autoware_ml/metrics/segmentation3d/partial_detection.py
```diff
@@ -0,0 +1,185 @@
...
+
+from autoware_ml.metrics.base import EvalStage, Metric
+from autoware_ml.metrics.detection3d.matching import mean_valid
+from autoware_ml.metrics.segmentation3d.point_cloud import (
+    PointCloudSegState,
+    valid_point_mask,
+)
+
+
+def _points_in_box(coord: np.ndarray, box: np.ndarray) -> np.ndarray:
```
mojomex: This is the same rule as `_count_points_in_rotated_boxes` in `autoware_ml/transforms/boxes3d/filters.py`. Consider exposing a shared point-in-box mask there instead of a second copy, so the two cannot drift.

autoware_ml/metrics/segmentation3d/partial_detection.py
```diff
@@ -0,0 +1,185 @@
...
+        if not self.class_names or len(set(self.class_names)) != len(self.class_names):
+            raise ValueError("class_names must name each scored detection class once.")
+        unknown = sorted(set(self.class_names) - set(self.det_class_names))
+        if unknown:
+            raise ValueError(f"class_names {unknown} are not detection classes.")
+        self.scored_labels: dict[int, str] = {
+            self.det_class_names.index(name): name for name in self.class_names
+        }
+
+    def _credit(self, k: int, n: int) -> float:
```
mojomex, line 92: This is the metric's defining formula, so give it a docstring (s(k)/s(n) with s(x) = x/(x+h)) and a nested `def` instead of a lambda with a noqa.

autoware_ml/metrics/segmentation3d/error_clusters.py
```diff
@@ -0,0 +1,109 @@
...
+)
+from autoware_ml.metrics.segmentation3d.spatial import cluster_sizes
+
+
+class ErrorClusters(Metric[PointCloudSegState]):
+    """Misclassified-point rate plus the count of error clusters."""
+
+    def __init__(
+        self,
+        cluster_radius: float = 0.5,
```
mojomex, line 42: Future work (not blocking): a fixed 0.5 m linking radius makes the cluster count range dependent. A 3 m phantom line sampled at 0.05 m spacing (about 5 m range) is 1 cluster, the same line at 0.6 m spacing (about 60 m range) is 5. The far range buckets will over-count phantoms; either scale the radius with range or note it in the docstring.

autoware_ml/metrics/segmentation3d/calibration.py
```diff
@@ -0,0 +1,92 @@
...
+import numpy as np
+
+from autoware_ml.metrics.base import EvalStage, Metric
+from autoware_ml.metrics.segmentation3d.point_cloud import (
+    PointCloudSegState,
+    valid_point_mask,
+)
+
+
+def _ece_from_bins(conf_sum: np.ndarray, correct_sum: np.ndarray, count: np.ndarray) -> float:
```
mojomex, line 21: Please add a one-line docstring (weighted sum over non-empty bins of |accuracy - mean confidence|, NaN when empty).

autoware_ml/metrics/segmentation3d/error_clusters.py
```diff
@@ -0,0 +1,109 @@
...
+            stage: Evaluation stage being reported.
+
+        Returns:
+            Metric keys mapped to values.
+        """
+        num_classes = state.num_classes
+        error_total = valid_total = cluster_total = num_frames = 0
+        per_error = [0] * num_classes
+        per_valid = [0] * num_classes
+        per_cluster = [0] * num_classes
```
mojomex: These names are not self-explanatory. Please either rename or add a one-line comment.

autoware_ml/metrics/segmentation3d/partial_detection.py
```diff
@@ -0,0 +1,185 @@
...
+class PartialDetectionScore(Metric[PointCloudSegState]):
+    """Saturating credit for partial segmentation of small-object detection boxes.
+
+    The metric joins two label spaces, the detection classes of the boxes and the
+    segmentation classes of the points, through one rule: the segmentation taxonomy
+    starts with the detection taxonomy, so a detection class and its segmentation
+    class share an index and no hand written mapping can drift. It refuses to run
+    on a state whose class set differs from ``seg_class_names``: a grouped suite
+    folds labels into another index space, where the shared index would silently
+    score the wrong class.
```
mojomex: This is a bit rambly, please make more concise and clear.

autoware_ml/metrics/segmentation3d/tolerant_error.py
```diff
@@ -0,0 +1,90 @@
...
+            state: Point-cache state for one (filter, range) bucket.
+            stage: Evaluation stage being reported.
+
+        Returns:
+            Metric keys mapped to values.
+        """
+        num_classes = state.num_classes
+        error_total = valid_total = 0
+        per_error = [0] * num_classes
+        per_valid = [0] * num_classes
```
mojomex: Please rename or clarify

mojomex (dismissed):
Apart from some small feedback, LGTM. Please address the feedback before merging.

autoware_ml/metrics/segmentation3d/error_clusters.py
amadeuszsz: Renamed them, `scored_points` for what the mask kept, `wrong_points` for prediction different than target, and the per class ones. Added a line above the block so you don't have to read the loop to find out.

Addressed in https://github.com/tier4/autoware-ml/pull/105/commits/a4d4bb6b219b77ce4060b6febfad87b9985d9597

autoware_ml/metrics/segmentation3d/tolerant_error.py
amadeuszsz: Same rename as in `error_clusters.py`, both metrics use the same words now.

Addressed in https://github.com/tier4/autoware-ml/pull/105/commits/a4d4bb6b219b77ce4060b6febfad87b9985d9597

autoware_ml/metrics/segmentation3d/partial_detection.py
amadeuszsz: Good catch, that was a real bug. Both counts now drop interior points whose target is not the box class, so ground returns can't dilute or inflate anything. Added your two cases as tests, all pedestrian points correct gives 1.0 and ground mispredicted as pedestrian gives 0.0.

Addressed in https://github.com/tier4/autoware-ml/pull/105/commits/0b60078cf29fb09db08f8f57ee63c0060b5e268c

autoware_ml/metrics/segmentation3d/partial_detection.py
amadeuszsz: Agreed. I put it in `geometry/utils.py` as `points_in_rotated_box` instead of in the filters, so metrics don't have to import from transforms. `_count_points_in_rotated_boxes` calls it now, so both sides share the rule.

Addressed in https://github.com/tier4/autoware-ml/pull/105/commits/2bb8d37098db864485a487c8a3c7ce3e556ec028

autoware_ml/metrics/segmentation3d/partial_detection.py
amadeuszsz: Documented, s(k)/s(n) with s(x) = x/(x+h), plus why it saturates and what h means. Lambda and the noqa are gone, nested def now.

Addressed in https://github.com/tier4/autoware-ml/pull/105/commits/7b585ddcd09707c37d2f9fff8a3dba17f04bb934

autoware_ml/metrics/segmentation3d/calibration.py
amadeuszsz: Added, count weighted sum over non empty bins of the gap between accuracy and mean confidence, NaN when every bin is empty.

Addressed in https://github.com/tier4/autoware-ml/pull/105/commits/9d192897bd62c8e51cf4e2b6fb46da495818848f

autoware_ml/metrics/segmentation3d/error_clusters.py
amadeuszsz: Noted it in the docstring, the count is range dependent and should be compared within a range bucket, not across. Left the radius fixed for now since you marked it non blocking, scaling it with range is worth its own change.

Addressed in https://github.com/tier4/autoware-ml/pull/105/commits/4ebb5fa933695a48eeaf02a17eb9c382d1470fb9

autoware_ml/metrics/segmentation3d/partial_detection.py
amadeuszsz: Cut it down. One line on what the score is, one paragraph on the shared index and why a grouped state is refused.

Addressed in https://github.com/tier4/autoware-ml/pull/105/commits/3d21642ba3112aa4b8c70b221e3b10f2d6485842

PR #106, opened 2026-08-25
feat(models): emit per-frame seg_frames eval output

amadeuszsz:
## Summary

All segmentation model paths emit the per-frame `seg_frames` eval output.

- FRNet and the PTv3 segmentation head (the joint segdet model delegates to it) switch `build_eval_output` from the flat point tensors to `segmentation_frames_eval_output` (#104), recovering each original point's frame from the batch `offset` and the sampled-space `inverse` map.
- Per-frame metadata and detection ground truth pass through into each frame entry when the batch carries them.
- The model unit tests move to the new contract: eval batches carry `offset` and assertions read the per-frame entries.

## Related Issues

Stacked on #105.

## Checklist

## Additional Notes

The bundled configs swap their metric suite target to the confusion suite here, in the same commit that changes the producer contract, and the retained range-only suite is removed with its last reference.

## Additional Log and Media

`pytest autoware_ml/tests/{metrics,configs,datamodule,models}`: 323 passed, 1 skipped at this branch head.


autoware_ml/models/segmentation3d/heads/ptv3.py
```diff
@@ -216,25 +220,36 @@ def prepare_for_export(self, order: Sequence[str]) -> PTv3SegDecoderHead:
...
     """
-    return {
-        "seg_pred_labels": seg_logits.argmax(dim=1)[batch["inverse"].long()],
-        "seg_target_labels": batch["origin_segment"].long(),
-        "seg_coord": batch["origin_coord"],
-    }
+    inverse = batch["inverse"].long()
+    offset = batch["offset"].long()
+    scores = torch.softmax(seg_logits, dim=1)[inverse]
+    return segmentation_frames_eval_output(
```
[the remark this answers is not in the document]
amadeuszsz: Fixed in the current head: this commit now also migrates both bundled dataset configs to the confusion suite (which consumes seg_frames), so the producer contract and its configured consumer change together and the range-only suite is removed with its last reference. The reviewed snapshot predates that force-push.

autoware_ml/metrics/detection3d/suite.py
```diff
@@ -79,8 +201,29 @@ def _warn_on_range_class_caps(self) -> None:
...
+        The collision TTC needs the ego pose and scene token exactly like the
+        region filters do. Declaring them here makes a missing key fail loud at
+        the first batch instead of silently reporting zero coverage.
+
+        Returns:
+            The required eval-output key names.
+        """
+        keys = super().required_keys()
+        if self.collision is not None:
+            keys = tuple(dict.fromkeys(keys + ("ego2global", "scene_token")))
```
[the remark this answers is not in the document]
amadeuszsz: Same stack split as on #102: the frame context is forwarded and collated in #107, the PR directly above this one (`detection_eval_output` plus the collation maps in the task base configs). This PR only switches the segmentation producers to seg_frames.

autoware_ml/models/segmentation3d/heads/ptv3.py
```diff
@@ -235,25 +239,36 @@ def prepare_for_export(self, order: Sequence[str]) -> PTv3SegDecoderHead:
...
+        ``{"seg_frames": [...]}`` keyed for the segmentation suites.
     """
-    return {
-        "seg_pred_labels": seg_logits.argmax(dim=1)[batch["inverse"].long()],
-        "seg_target_labels": batch["origin_segment"].long(),
-        "seg_coord": batch["origin_coord"],
-    }
+    inverse = batch["inverse"].long()
+    offset = batch["offset"].long()
+    scores = torch.softmax(seg_logits, dim=1)[inverse]
```
mojomex, line 262: This materialises an (N_original, C) softmax on every val step, but the confusion suite never reads `scores`; only the point suites (PR 105) do. Consider producing scores only when a consumer is configured, or gathering logits and letting the suite softmax what it needs.

autoware_ml/models/segmentation3d/frnet.py
```diff
@@ -224,14 +238,32 @@ def compute_metrics(
...
+        Returns:
+            Flat dict with the per-frame ``seg_frames`` the suites read.
+        """
         point_logits = outputs[0]
-        return {
-            "seg_pred_labels": point_logits.argmax(dim=1),
-            "seg_target_labels": batch["pts_semantic_mask"],
-            "seg_coord": batch["points"][:, :3],
-        }
+        offset = batch["offset"].long()
```
mojomex: `offset` is generated from whichever key is the first `concat` entry in `collation_map`; here that is `points`, which matches `point_logits` rows. Worth one line in the docstring, since reordering the YAML silently changes which space the offsets describe.

mojomex:
Unrelated commits 3fee951, 3d0d10f seem to have buried the PR's actual commit, please check.

mojomex (dismissed):

autoware_ml/models/segmentation3d/heads/ptv3.py
amadeuszsz: Right, val never reads scores. Same with `ego2global` and `scene_token`, builders attach what the batch happens to have, metrics cannot request what they need. Tbh this needs a significant refactor, it touches both task builders and the signature every model implements. I will address it in a follow-up PR.

PR #107, opened 2026-08-25
feat(datamodule): per-frame evaluation metadata for the map filters

amadeuszsz:
## Summary

Datamodules expose the per-frame evaluation metadata the map-backed filters and the collision provider require.

- `frame_meta` derives the scene directory fragment and per-frame `ego2global` and `scene_token`, shared by the T4 detection, segmentation, segdet and multiview datamodules.
- `detection_eval_output` forwards `gt_num_points`, `ego2global` and `scene_token` when the batch carries them, closing the gap between the detection suites' `required_keys()` and the built-in models.
- The segmentation datamodule batches carry the metadata into each `seg_frames` entry through the builder added earlier in the stack.
- The segmentation PTv3 base task config collates the new keys.

## Related Issues

Stacked on #106.

## Checklist

## Additional Notes

With this PR every producer-side piece is in place, the config migration PR next in the stack turns the filters and collision metrics on.

## Additional Log and Media

`pytest autoware_ml/tests/{metrics,configs,datamodule,models}`: 328 passed, 1 skipped at this branch head.


autoware_ml/datamodule/common/frame_meta.py
```diff
@@ -0,0 +1,29 @@
...
+    annotation token, it lets :class:`T4LaneletMapResolver` locate the scene's
+    lanelet map directly under ``data_root``.
+    """
+    parts = PurePosixPath(lidar_path).parts
+    if len(parts) < 4:
+        raise ValueError(
+            f"Cannot derive a scene directory from lidar path {lidar_path!r}; expected "
+            "'<db>/<scene_uuid>/<version>/...'."
+        )
+    return "/".join(parts[:3])
```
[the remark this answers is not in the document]
amadeuszsz: Confirmed, the fragment derivation ignored the stored-path forms the path resolver supports. scene_dir_fragment now takes data_root, strips it when the stored path is prefixed by or absolute under it, and raises for an absolute path outside it (such a scene could never resolve a map and a nonsense token would silently merge scenes in the cache and coverage). Tests cover all three forms.

autoware_ml/metrics/segmentation3d/point_cloud.py
```diff
@@ -0,0 +1,492 @@
...
+        boxes_np = boxes.numpy()
+        labels_np = labels.numpy()
+
+        keep = np.ones(coord_np.shape[0], dtype=bool) if region_mask is None else region_mask
+        if metric_range is not None:
+            keep = keep & self._range_mask(coord_np[:, :2], metric_range)
+        # Boxes carry no region membership of their own, clip only by range.
+        if metric_range is not None and boxes_np.shape[0]:
+            box_keep = self._range_mask(boxes_np[:, :2], metric_range)
+            boxes_np, labels_np = boxes_np[box_keep], labels_np[box_keep]
```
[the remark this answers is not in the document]
amadeuszsz: Valid, boxes bypassed the filter axis. The point-cloud suite now resolves per-box filter membership at update through the same filters (footprint overlap, zeroed for uncovered scenes) and state_for clips the auxiliary boxes by the filter column alongside the radial range. Fixed in the suite's own PR (#104) with a regression test asserting an off-region box is omitted from the filtered state.

autoware_ml/datamodule/nuscenes/multiview_detection3d.py
```diff
@@ -36,6 +38,21 @@ class NuscenesMultiviewDetection3DDataset(MultiviewDetection3DDataset):
...
+        Args:
+            index: Dataset sample index.
+
+        Returns:
+            Metadata dictionary with ``ego2global`` attached, the pose the
+            evaluation filters read. The scene token comes from the common
+            record.
+        """
+        data_info = super().get_data_info(index)
+        data_info["ego2global"] = np.asarray(self.data_infos[index]["ego2global"], dtype=np.float64)
```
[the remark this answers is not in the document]
amadeuszsz: Right, nuScenes records also store the pose as translation plus rotation. The override now takes the matrix the common record already assembles and raises when a sample carries no pose at all, with a test using the component form. Addressed in https://github.com/tier4/autoware-ml/pull/107/commits/7d59d34d5bede73f55b1fb1a2e2eaf26fe839446.

autoware_ml/metrics/segmentation3d/partial_detection.py
```diff
@@ -0,0 +1,173 @@
...
+            boxes = [
+                (box, int(box_label))
+                for box, box_label in zip(frame.gt_boxes, frame.gt_box_labels, strict=True)
+                if int(box_label) in self.box_label_to_seg_class
+            ]
+            if not boxes:
+                continue
+            valid = valid_point_mask(frame, state.num_classes, state.ignore_index)
+            coord_xy = frame.coord[valid][:, :2].astype(np.float64)
+            pred = frame.pred[valid]
```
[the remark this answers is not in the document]
amadeuszsz: Valid, membership was BEV-only so a return above the box changed n, the low-point skip and the credit. It now bounds by the box height like the annotation point counter does. Addressed in the metric's own PR (#105), https://github.com/tier4/autoware-ml/pull/105/commits/f2d61d5afeb8782d6d492c0e195fdc69332b8e85.

autoware_ml/metrics/segmentation3d/partial_detection.py
```diff
@@ -0,0 +1,173 @@
...
+            half_saturation: Correct-point count earning half the existence credit.
+            min_points: Smallest interior point count a box needs to be scored.
+            stages: Stage names this metric reports for, as in :class:`Metric`.
+            filter: Optional selection axis, as in :class:`Metric`.
+        """
+        super().__init__(stages, filter=filter)
+        self.box_label_to_seg_class = {int(k): int(v) for k, v in box_label_to_seg_class.items()}
+        self.seg_class_names = tuple(seg_class_names)
+        self.class_names = tuple(class_names) if class_names is not None else None
+        self.half_saturation = float(half_saturation)
```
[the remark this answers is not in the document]
amadeuszsz: Agreed, h = 0 divides by zero for a box without a single correct point. Rejected at construction next to min_points. Addressed in the metric's own PR (#105), https://github.com/tier4/autoware-ml/pull/105/commits/f2d61d5afeb8782d6d492c0e195fdc69332b8e85.

autoware_ml/datamodule/nuscenes/detection3d.py
```diff
@@ -129,6 +130,8 @@ def get_data_info(self, index: int) -> dict[str, Any]:
                 sample.get("num_features", sample.get("lidar_points", {}).get("num_pts_feats", 5))
             ),
             "sweeps": self._resolve_sweeps(sample),
+            "ego2global": np.asarray(sample["ego2global"], dtype=np.float64),
```
mojomex: Frame mismatch: nuScenes points and GT boxes stay in the LIDAR_TOP frame (nothing applies `lidar2ego`, which has a ~-90° yaw and 0.94 m offset), while the filters treat coordinates as base_link and transform with `ego2global` alone. It is harmless today only because nuScenes has no lanelet map, so the filters report unavailable. Either attach `ego2global @ lidar2ego` for nuScenes (and name the key for what it is), or state that the metadata is only valid where `lidar2ego` is identity (T4 LIDAR_CONCAT). Same applies to the nuscenes seg/segdet datasets.

autoware_ml/datamodule/t4dataset/segmentation3d.py
```diff
@@ -170,10 +173,17 @@ def get_data_info(self, index: int) -> dict[str, Any]:
...
-            return self._get_per_sensor_info(sample, sample_index, source_name)
-
-        sample = self.data_infos[index]
-        return self._get_sample_info(sample)
+            info = self._get_per_sensor_info(sample, sample_index, source_name)
+        else:
+            sample = self.data_infos[index]
+            info = self._get_sample_info(sample)
+
+        info["ego2global"] = np.asarray(sample["ego2global"], dtype=np.float64)
```
mojomex: This runs on the training path too, so info files without `ego2global`/`timestamp` now fail at the first sample instead of at eval. The configured `info/lidarseg/t4dataset_j6gen2_*` pkls carry both, but e.g. `calibration_info_lidarseg/t4dataset_gen2_lidarseg_infos_*.pkl` on the NAS does not. Fine if intended, but then the PR should say infos need regenerating; otherwise attach only when present and let the suite fail loud, as `eval_output.py` promises.

autoware_ml/datamodule/common/tests/test_multiview_detection3d.py
```diff
@@ -84,6 +88,87 @@ def test_multiview_detection_dataset_applies_loader_pipeline(tmp_path: Path) ->
     assert output["gt_labels"].tolist() == [0]
 
 
+def test_nuscenes_multiview_exposes_the_ego_pose_from_either_record_form(
```
mojomex: What two record forms is this name referring to?

autoware_ml/datamodule/t4dataset/frame_meta.py
```diff
@@ -0,0 +1,47 @@
...
+datasets attach both at metadata time and the eval-output builders pass them
+through per frame.
+"""
+
+from __future__ import annotations
+
+from pathlib import PurePosixPath
+
+
+def scene_dir_fragment(lidar_path: str, data_root: str) -> str:
```
mojomex, line 16: This seems to be t4dataset-specific. Please consider moving it out of `common`.

mojomex (dismissed):
Almost LGTM, please check the comments.

autoware_ml/datamodule/nuscenes/detection3d.py
amadeuszsz: You're right, nothing applies `lidar2ego` on the nuScenes side, so the pose I attached was the wrong frame. It only goes unnoticed because there is no lanelet map and the filters skip those scenes. The infos already carry `lidar2ego`, so all four nuScenes datasets compose it through one `lidar_to_map` helper now and `ego2global` means sensor frame to map everywhere. It raises when the mounting is missing rather than silently using the ego pose.

Addressed in https://github.com/tier4/autoware-ml/pull/107/commits/3f1b201b06542b6d1fb9e5eb43e1fe3b033de66d

autoware_ml/datamodule/common/tests/test_multiview_detection3d.py
amadeuszsz: Translation plus rotation, or a ready 4x4 matrix. The name made you go looking for that, so it is two tests now, one per form. The matrix case was not actually covered before either, despite the name.

Addressed in https://github.com/tier4/autoware-ml/pull/107/commits/438a20bd1453531cc5a581960c3f3e58d84bf9d2

autoware_ml/datamodule/t4dataset/frame_meta.py
amadeuszsz: Yes, it parses the T4 scene layout and only t4 modules call it. Moved to `datamodule/t4dataset/frame_meta.py`.

Addressed in https://github.com/tier4/autoware-ml/pull/107/commits/7c9dab66ad0d9ded15c34e93722771fdbb5bb6c4

PR #113, opened 2026-09-02
feat(ptv3)!: name serialization tensors after levels, not pooling stages

mojomex:
## Status quo

The exported graph names a level's tensors after the pooling stage that produced them. Level `l`'s
coordinates and serialization order are `serialized_pooling_{l-1}_grid_coord`,
`_serialized_order` and `_serialized_inverse`, while level 0's are unprefixed. Every consumer
carries that off-by-one, and the detection head spells the same tensor a third way:
`point_grid_coord_{l}`. So level 3's coordinates are `serialized_pooling_2_grid_coord` in the
encoder and the segmentation head, and `point_grid_coord_3` in the detection head — one tensor,
two names, same model.

The dynamic axes have a sharper version of the same problem. A stage's input count and the
previous stage's output count are the *same number*, but they are two unrelated `dim_param`s:

- `num_voxels` and `serialized_pooling_0_in_voxels` both mean level 0
- `serialized_pooling_{i}_out_voxels` and `serialized_pooling_{i+1}_in_voxels` both mean level `i+1`

It shows up inside a single file: in the detection head, `point_feat_3` is
`['serialized_pooling_2_out_voxels', 256]` while `pooling_cluster_3` is
`['serialized_pooling_3_in_voxels']`. Shape inference cannot relate the two, so an artifact whose
names disagreed would build and run unchallenged.

## What this does

A **level** is a resolution; a **pooling stage** is the transition between two levels, so `S`
levels have `S - 1` stages. Tensors are named after whichever they belong to.

The three `SerializedPoolingMeta` fields that describe a stage's *output level* move to the level:

| before                                       | after                        |
| -------------------------------------------- | ---------------------------- |
| `grid_coord`                                 | `level_0_grid_coord`         |
| `serialized_order`                           | `level_0_serialized_order`   |
| `serialized_inverse`                         | `level_0_serialized_inverse` |
| `serialized_pooling_{i}_grid_coord`          | `level_{i+1}_grid_coord`     |
| `serialized_pooling_{i}_serialized_order`    | `level_{i+1}_serialized_order`   |
| `serialized_pooling_{i}_serialized_inverse`  | `level_{i+1}_serialized_inverse` |
| `point_grid_coord_{l}` (detection head)      | `level_{l}_grid_coord`       |

`indices`, `indptr`, `cluster` and `head_indices` describe the transition, not a level, and keep
`serialized_pooling_{i}_`. `indices` is level-`i`-sized while `indptr` is level-`i+1`-sized plus
one, so they genuinely span the transition. `feat` is unprefixed because only level 0 has one, and
`point_feat_{l}` / `pooling_cluster_{i}` are already correct.

Dynamic axes become one `level_{l}_voxels` symbol per level. Exporting the encoder, its distinct
symbolic dimensions go from **13 to 9**, the remaining four being the `level_{l}_voxels_plus_one`
that `indptr` needs because a `dim_param` cannot express arithmetic.

The payoff in the exporter is that level 0 stops being a special case:
`seg_head_export_input_names` loses its `prefix = "" if stage == 0 else ...` branch, and
`build_seg_head_input_dynamic_axes` loses the matching `if stage == 0` block. Naming lives in three
one-line helpers — `level_tensor_name`, `level_voxel_axis_name`, `pooling_tensor_name` — plus
`pooling_field_dynamic_axis`, which replaces a function that parsed tensor names back into
`(stage, field)` pairs.

## Related links

**Parent Issue:**

- None

## How was this PR tested?

- **Real export, names read back from the artifacts.** Exported `segmentation3d/ptv3/voxel012_122m_t4dataset_j6gen2`
  and inspected both graphs. The encoder declares 28 inputs in the intended scheme
  (`level_0_grid_coord`, `feat`, `level_0_serialized_order`, `level_0_serialized_inverse`, then per
  stage `serialized_pooling_{i}_{indices,indptr,head_indices}` and per level
  `level_{i+1}_{grid_coord,serialized_order,serialized_inverse}`), and the segmentation head
  declares `level_2_*` and `level_3_*` for `dec_depths: [0, 0, 1, 1]` — levels 2 and 3, which is
  what the rule says. Symbolic dimensions: 13 → 9 in the encoder, with `level_0_voxels` now used
  5× where `num_voxels` and `serialized_pooling_0_in_voxels` previously split those uses.
- **Full test suite**: 503 passed. `autoware_ml/tests/models/test_bevfusion.py` and
  `autoware_ml/tests/ops/test_bev_pool.py` fail to collect on this image before and after this
  change (`bev_pool_ext` is not built), so they are excluded.
- **New coverage**: `test_det_head_export_names_address_levels_not_pooling_stages` pins the
  detection head's input-name and dynamic-axis rule, which had none — and which is where the
  third spelling of level `l`'s coordinates lived.
- `pre-commit` clean.

## Notes for reviewers

The rename was applied by asking, per tensor, whether its index runs `0..S-1` (level) or `0..S-2`
(stage). Watch for the distinction between ONNX names and `Point`/batch dictionary keys: `grid_coord`,
`serialized_order` and `serialized_inverse` remain dictionary keys inside the model and the
transforms, and are untouched there. Only the exported graph's names change.

## Stack

- **#85** (`feat/ptv3-level0-order-inputs`): takes the level-0 serialization order as a graph input.
  **Must merge first** — this PR is branched from it, so the diff shown here includes it. Review
  only the top commit, or compare against #85's head.
- **This PR**: the naming change.

It is worth doing now rather than later: #85 already forces a re-export of every artifact, so
landing this in the same breaking change costs no additional re-export cycle. That window closes
when #85 merges.

The matching deployment change in `autoware_universe` is not yet open; the node currently declares
the old names and will need `level_{l}_*` before either can be deployed. autowarefoundation/autoware_universe#13155
is the corresponding node-side PR for #85, and autowarefoundation/autoware_universe#13307 lands the
same level/stage vocabulary in the node's C++ ahead of it.

## Interface changes

15 graph input names and 9 dynamic-axis names change; `point_grid_coord_{l}` is removed in favour
of `level_{l}_grid_coord`.

## Effects on system behavior

None on the computed graph — the tensors, their shapes, dtypes and order are unchanged. Existing
ONNX artifacts and TensorRT engines must be re-exported, and the deploying node must declare the
new names.


mojomex:
Closing: the level-based tensor naming is too big a change for the gain, and it would force a re-export on everyone. Leaving the exported names as-is.

PR #116, opened 2026-09-03
feat(datamodule): add scene-streaming dataloaders for temporal detectors

vividf:
## Summary
Adds the scene-streaming data pipeline that stateful temporal detectors (StreamPETR) require: every dataloader lane must walk one scene frame by frame so the model's memory bank sees contiguous streams.

## Changes
- **`GroupStreamingSampler`** (`datamodule/samplers.py`): assigns whole scenes to batch lanes, so across consecutive batches, position *k* always continues the same scene. Shards scenes by rank for DDP (every rank derives all ranks' lists from the same seed+epoch, then every list is truncated to the shortest one so all ranks iterate the same number of steps and the collectives cannot deadlock), reshuffles scene order per epoch via `set_epoch`, and trims every rank to whole batches — the trim is reported at WARNING level. The trainer must run with `use_distributed_sampler: false` (the StreamPETR base config sets it; documented on the class): Lightning would otherwise wrap the sampler and destroy the lane contiguity.
- **`MultiviewDetection3DDataModule`** (`datamodule/common/multiview_detection3d.py`): shared scaffolding that wires the streaming sampler into every split when `streaming: true`. Dataloader-level `shuffle` is rejected — the sampler owns the sample order. The NuScenes/T4 datamodules shrink to thin `_create_dataset` adapters. `streaming` defaults to `false` because BEVFusion shares these datamodules and is not temporal.
- Collation keeps numeric numpy scalars' dtype (float64 epoch-second timestamps would be quantized to 256 s steps by float32); non-numeric numpy scalars pass through unchanged.
- Datamodule and sampler unit tests move next to the modules they cover, per review. CI and the pixi `test` task now run `pytest autoware_ml/ -v` instead of `pytest autoware_ml/tests/ -v`, so in-module tests are actually collected. This also picks up three pre-existing in-module test files that were never executed (`models/detection3d/task_modules/tests`, `ops/voxelization/tests`, `preprocessing/detection3d/tests`).

## Notes for reviewers
- Evaluation configs deliberately use `batch_size: 1` per rank (one lane) so no eval frame is trimmed; the trim warning makes any residual drop visible.
- The per-epoch scene shuffle is seeded by `seed + epoch` (default seed 0, matching the accepted production runs).
- The `ResizeCropFlipRotImage` intrinsics fix that was originally in this PR moved out to #117.

## Testing
- `datamodule/tests/test_samplers.py`: lane continuity across batches, whole-batch trimming, cross-rank length agreement + disjoint shards, deterministic `set_epoch` reshuffle, unfillable-lane error, dataloader-shuffle rejection, trim warning.
- `datamodule/common/tests/test_multiview_detection3d.py`: prev-frame continuity flags, camera-order frame filtering, float64 timestamp collation.

**Stack 1/9** — base: `main`.

## Stack
1. **feat(datamodule): add scene-streaming dataloaders for temporal detectors ← this PR**
2. fix(transforms): fold ResizeCropFlipRotImage crop into the aug matrix exactly — #117
3. feat(models): add StreamPETR building blocks — #118
4. feat(models): add the StreamPETR detector — [vividf#9](https://github.com/vividf/autoware-ml/pull/9)
5. feat(streampetr): add base nuScenes and T4 j6gen2 configurations — [vividf#10](https://github.com/vividf/autoware-ml/pull/10)
6. feat(streampetr): harden the streaming training infrastructure — [vividf#3](https://github.com/vividf/autoware-ml/pull/3)
7. feat(streampetr): project 2D annotations and fold augmentations into ego poses — [vividf#4](https://github.com/vividf/autoware-ml/pull/4)
8. feat(streampetr): add auxiliary 2D head, partial-ignore, and DDP loss normalization — [vividf#5](https://github.com/vividf/autoware-ml/pull/5)
9. feat(streampetr): pin the production three-stage training configs — [vividf#6](https://github.com/vividf/autoware-ml/pull/6)

#116, #117 and #118 are independent of each other and all based on `main`, so they can be reviewed and merged in any order. Each later layer is opened here once its prerequisites merge, so every PR stays a single-layer diff.


autoware_ml/datamodule/common/multiview_detection3d.py
```diff
@@ -27,17 +27,124 @@
...
 )
 from autoware_ml.datamodule.common.serialization import SerializedSampleList
+from autoware_ml.datamodule.samplers import GroupStreamingSampler
 
 logger = logging.getLogger(__name__)
 
 
+def build_streaming_dataloader(
+    dataset: Dataset,
+    dataloader_cfg: Any,
```
KSeangTan: Please avoid using `Any` since it doesn't bring any useful hints.

autoware_ml/datamodule/common/multiview_detection3d.py
```diff
@@ -27,17 +27,124 @@
...
+    Args:
+        dataset: Dataset exposing ``scene_index_groups()``.
+        dataloader_cfg: Split dataloader configuration.
+        collate_fn: Batch collation callable.
+        shuffle_scenes: Whether the scene order is reshuffled every epoch.
+
+    Returns:
+        Streaming dataloader for the split.
+    """
+    dataloader_kwargs = dataloader_cfg.to_dataloader_kwargs()
```
KSeangTan: Do not use any `kwargs`, it is hard to read and understand.

autoware_ml/datamodule/common/multiview_detection3d.py
```diff
@@ -27,17 +27,124 @@
...
+    if dataloader_kwargs.pop("shuffle"):
+        raise ValueError(
+            "Streaming datamodules own the sample order; set dataloader shuffle to false."
+        )
+    sampler = GroupStreamingSampler(
+        dataset, batch_size=dataloader_cfg.batch_size, shuffle=shuffle_scenes
+    )
+    return DataLoader(dataset=dataset, sampler=sampler, collate_fn=collate_fn, **dataloader_kwargs)
```
KSeangTan: No `**kwargs` when passing, only passing parameters explicitly.

autoware_ml/datamodule/common/multiview_detection3d.py
```diff
@@ -27,17 +27,124 @@
...
+    behaves identically.
+    """
+
+    def __init__(
+        self,
+        data_root: str,
+        train_ann_file: str,
+        val_ann_file: str,
+        test_ann_file: str,
+        class_names: list[str],
```
KSeangTan: use `Sequence` instead of `list`.

autoware_ml/datamodule/common/multiview_detection3d.py
```diff
@@ -27,17 +27,124 @@
...
+        camera_order: list[str],
+        name_mapping: dict[str, str] | None = None,
```
KSeangTan: use `MappingProxyType` for immutability.

autoware_ml/datamodule/common/multiview_detection3d.py
```diff
@@ -27,17 +27,124 @@
...
+        filter_frames_with_camera_order: bool = True,
+        require_image_files: bool = False,
+        streaming: bool = False,
+        **kwargs: Any,
```
KSeangTan: No `kwargs`.

autoware_ml/datamodule/common/multiview_detection3d.py
```diff
@@ -27,17 +27,124 @@
...
+            val_ann_file: Validation annotation file path.
+            test_ann_file: Test annotation file path.
+            class_names: Ordered detector class names.
+            camera_order: Ordered camera names expected by the model.
+            name_mapping: Optional mapping from dataset labels to detector labels.
+            filter_frames_with_camera_order: Drop frames missing any camera in
+                ``camera_order`` so image loading always sees a complete set.
+            require_image_files: Also verify camera image files exist on disk
+                while filtering (slower; off by default).
+            streaming: Whether every dataloader streams scene-contiguous
```
KSeangTan: I think streaming should always be true when we are working for streaming/temporal models?

autoware_ml/datamodule/common/multiview_detection3d.py
```diff
@@ -27,17 +27,124 @@
...
+        super().__init__(**kwargs)
+        self.data_root = data_root
+        self.class_names = class_names
+        self.camera_order = camera_order
+        self.name_mapping = {} if name_mapping is None else dict(name_mapping)
+        self.filter_frames_with_camera_order = filter_frames_with_camera_order
+        self.require_image_files = require_image_files
+        self.streaming = streaming
+
+        def resolve_ann_file(ann_file: str) -> str:
```
KSeangTan: No nested functions.

autoware_ml/datamodule/common/multiview_detection3d.py
```diff
@@ -27,17 +27,124 @@
...
+            "test": resolve_ann_file(test_ann_file),
+            "predict": resolve_ann_file(test_ann_file),
+        }
+
+    def _create_dataloader(self, split: str):
+        """Create the split dataloader, streaming scenes on every split."""
+        if not self.streaming:
+            return super()._create_dataloader(split)
+        return build_streaming_dataloader(
+            getattr(self, f"{split}_dataset"),
```
KSeangTan, line 193: If we cannot ensure there's variables: `{split}_dataset` and `{split}_dataloader_cfg` in the class, and need to use `getattr`, i think it's a bad design.

autoware_ml/datamodule/common/multiview_detection3d.py
```diff
@@ -285,6 +392,7 @@ def get_data_info(self, index: int) -> dict[str, Any]:
         sample = self.data_infos[index]
         lidar_path = self._resolve_lidar_path(sample)
         camera_infos = self._get_camera_infos(sample)
+        timestamp = sample.get("timestamp")
```
KSeangTan: timestamp = sample.get("timestamp", None)

autoware_ml/datamodule/samplers.py
```diff
@@ -77,3 +80,113 @@ def __iter__(self):
...
+class GroupStreamingSampler(Sampler[int]):
+    """Interleave scene-contiguous frame indices across dataloader lanes.
+
+    Each of the ``batch_size`` dataloader lanes walks whole scenes frame by
+    frame, so consecutive iterations feed consecutive frames of the same scene
+    into the same batch position. Stateful temporal models can therefore carry
+    per-lane memory across iterations, with scene changes signalled by the
+    dataset's ``prev_exists`` metadata. Scenes are shuffled per epoch,
+    partitioned round-robin over distributed ranks and lanes, and every rank
+    is trimmed to the shortest rank so all batches stay complete.
```
KSeangTan: what does it mean by `shortest rank`?

autoware_ml/datamodule/common/tests/test_multiview_detection3d.py
```diff
@@ -126,3 +128,37 @@ def test_multiview_detection_dataset_builds_prev_exists_from_scene_tokens(
     assert dataset.get_data_info(0)["prev_exists"] == np.float32(0.0)
     assert dataset.get_data_info(1)["prev_exists"] == np.float32(1.0)
     assert dataset.get_data_info(2)["prev_exists"] == np.float32(0.0)
+
+
+def test_multiview_detection_dataset_keeps_timestamp_float64(tmp_path: Path) -> None:
```
KSeangTan, line 133: We start moving unit tests to their respective folder instead, for example,
`common/tests/test_multiview_detection3d.py`

autoware_ml/transforms/camera/resize.py
```diff
@@ -223,44 +242,42 @@ def _augment_image(self, image: npt.NDArray) -> tuple[npt.NDArray[np.float32], n
         final_height, final_width = self.data_aug_conf["final_dim"]
         resize_lim = self.data_aug_conf["resize_lim"]
         bot_pct_lim = self.data_aug_conf["bot_pct_lim"]
+        # The base scale covers the final crop on both axes so one resize plus
```
KSeangTan: Move changes related to `crop` to another PR.

.cspell.json
```diff
@@ -123,11 +124,13 @@
     "getitem",
     "gloo",
     "graphsurgeon",
+    "Hardsigmoid",
```
KSeangTan: is this used anywhere?

.cspell.json
```diff
@@ -123,11 +124,13 @@
...
     "Hardswish",
     "hilberts",
     "hparams",
     "hsigmoid",
     "hyperparams",
+    "iccv",
```
KSeangTan: is this used anywhere?

.cspell.json
```diff
@@ -151,8 +154,10 @@
     "lovasz",
     "LSSFPN",
     "lucide",
+    "matmuls",
```
KSeangTan: is this used anywhere?

KSeangTan (changes requested):
Please address them accordingly

KSeangTan:
@codex review

autoware_ml/tests/datamodule/test_multiview_detection3d.py
```diff
@@ -126,3 +128,37 @@ def test_multiview_detection_dataset_builds_prev_exists_from_scene_tokens(
     assert dataset.get_data_info(0)["prev_exists"] == np.float32(0.0)
     assert dataset.get_data_info(1)["prev_exists"] == np.float32(1.0)
     assert dataset.get_data_info(2)["prev_exists"] == np.float32(0.0)
+
+
+def test_multiview_detection_dataset_keeps_timestamp_float64(tmp_path: Path) -> None:
```
KSeangTan: Also, please start using `unittest` framework from Python instead.

autoware_ml/datamodule/common/multiview_detection3d.py
vividf: Fixed: `dataloader_cfg: DataLoaderConfig` and `collate_fn: Callable[[list[dict[str, Any]]], dict[str, Any]]`.

autoware_ml/datamodule/common/multiview_detection3d.py
vividf: Fixed: it now reads `dataloader_cfg.shuffle` directly and passes each `DataLoader` argument explicitly.

For context, the same pattern is already merged in `build_detection_dataloader` (`datamodule/common/detection3d.py`) and `DataModule._create_dataloader` (`datamodule/base.py`) — this function was copied from them. I left those alone to keep this PR scoped, but happy to clean them up in a follow-up.

autoware_ml/datamodule/common/multiview_detection3d.py
vividf: Fixed: all six `DataLoader` arguments are passed explicitly now.

autoware_ml/datamodule/common/multiview_detection3d.py
vividf: Fixed: `class_names` and `camera_order` are `Sequence[str]`, copied into lists when stored.

autoware_ml/datamodule/common/multiview_detection3d.py
vividf: Stored as `MappingProxyType` now.

I kept the parameter annotated `Mapping[str, str] | None` rather than `MappingProxyType`, because Hydra passes a plain dict/DictConfig from YAML and cannot construct a `MappingProxyType`. So the immutability lives on the stored attribute.

autoware_ml/datamodule/common/multiview_detection3d.py
vividf: Fixed: `**kwargs` is gone, the nine base datamodule parameters are listed explicitly and forwarded by name.

I checked all four configs that instantiate this datamodule (two BEVFusion, two StreamPETR) — every key they pass is accepted by the new signature, so no config changes were needed.

autoware_ml/datamodule/common/multiview_detection3d.py
vividf: It has to stay a flag, because this datamodule is shared.

BEVFusion uses the same `Nuscenes`/`T4MultiviewDetection3DDataModule` subclasses (`bevfusion/camera_lidar_voxel0075_second_secfpn_54m_nuscenes.yaml` and the j6gen2 one), and it is not a temporal model, so it needs normal shuffling. The default is `false` and only the StreamPETR configs set it `true`. I expanded the docstring to say this explicitly.

autoware_ml/datamodule/common/multiview_detection3d.py
vividf: Fixed: moved to a module-level `_resolve_ann_file(data_root, ann_file)`.

autoware_ml/datamodule/common/multiview_detection3d.py
vividf: Agreed in principle, but the split-keyed `getattr` is the base class's existing dispatch, not something this PR introduces: `DataModule.setup()` and `DataModule._create_dataloader()` both use it, and three already-merged subclasses override `_create_dataloader(self, split)` the same way — `nuscenes/segdet.py:207`, `t4dataset/detection3d.py:349`, `t4dataset/segdet.py:210`.

Fixing it properly means changing the base signature to pass the dataset and config explicitly, which also touches those three overrides, `build_detection_dataloader`, and `test_base.py` — i.e. the CenterPoint/TransFusion/segdet paths. I'd rather not bundle that cross-cutting refactor into this PR, so I kept this override consistent with the base. Happy to open a dedicated refactor PR once this lands.

autoware_ml/datamodule/common/multiview_detection3d.py
vividf: Fixed.

autoware_ml/datamodule/samplers.py
vividf: Poor wording, reworded. Each rank builds its own index list, and every list is then truncated to the length of the shortest one, so all ranks iterate the same number of steps. Without that equalization the ranks would run a different number of steps and the collectives would deadlock.

autoware_ml/datamodule/common/tests/test_multiview_detection3d.py
vividf: Done: `test_samplers.py` -> `autoware_ml/datamodule/tests/`, `test_multiview_detection3d.py` -> `autoware_ml/datamodule/common/tests/`.

autoware_ml/transforms/camera/resize.py
vividf: Done, split out into #117.

That PR also documents the BEVFusion impact, since `ImageAug3D` reaches this transform — worth a look, as the augmented pixels change for configs whose `resize_lim` triggered the old hidden second resize.

.cspell.json
vividf: Correct, none of these are used in this PR — they belong to later layers of the stack. I removed all five I had added (`autocast`, `Hardsigmoid`, `iccv`, `matmuls`, `metas`); each one goes back in the layer that introduces its usage. `.cspell.json` is now unchanged by this PR.

.cspell.json
vividf: Same, removed. `iccv` belongs to the layer adding `docs/models/streampetr.md`.

.cspell.json
vividf: Same, removed. `matmuls` belongs to the layer adding the StreamPETR model tests.

vividf:
Thanks for the review — all comments addressed and pushed.

Two that I answered rather than changed:

- **`getattr(self, f"{split}_dataset")`**: kept as-is. It is the base class's existing split dispatch (`DataModule.setup()`, `DataModule._create_dataloader()`, plus three already-merged overrides in `nuscenes/segdet.py`, `t4dataset/detection3d.py`, `t4dataset/segdet.py`). Changing it properly means changing the base signature and those three overrides, which reaches the CenterPoint/TransFusion/segdet paths — I'd rather do that as a dedicated refactor PR after this lands than bundle it here.
- **`streaming` flag**: has to stay a flag, since BEVFusion shares these datamodules and is not temporal. Default `false`, only StreamPETR configs set it `true`.

The `ResizeCropFlipRotImage` change is now #117, and `.cspell.json` is untouched by this PR — all five words I had added belong to later layers.

One change worth a second look: moving the tests out of `autoware_ml/tests/` would have silently stopped them running, since CI only collected that directory, so CI and the pixi `test` task now run `pytest autoware_ml/ -v`. That also starts executing three in-module test files that existed but were never run. Let me know if you'd prefer to scope that differently.

autoware_ml/datamodule/common/multiview_detection3d.py
KSeangTan: IMO, i would rather to make it another `DataModule` for two different cases, but we should further refactor it once fully switch to `MultiTask`.

KSeangTan (dismissed):
Thanks, LGTM overall

vividf:
Branch updated (force-push) with one small fix on top of the review round: `autoware_ml/datamodule/__init__.py` and `autoware_ml/datamodule/common/__init__.py` were missing.

Moving the tests next to their modules (as requested) created `autoware_ml/datamodule/tests/` and `autoware_ml/datamodule/common/tests/`, but `datamodule/` was the only source package in the repo without an `__init__.py`. Without it pytest resolves both new test packages as top-level `tests.*`, which collides with `autoware_ml/tests/` and breaks collection of the whole tree:

```
ModuleNotFoundError: No module named 'tests.test_samplers'
ModuleNotFoundError: No module named 'tests.test_multiview_detection3d'
```

That matters because this PR also switches CI to `pytest autoware_ml/ -v`. With the two `__init__.py` files added, the full tree collects and passes (592 tests at this commit). `models/**/tests/` already worked because those parents are regular packages.

Also opened the next independent layer for review: #118 (StreamPETR building blocks — models only, independent of this PR).

KSeangTan:
@codex review

amadeuszsz (approved):

PR #117, opened 2026-09-04
fix(transforms): fold ResizeCropFlipRotImage crop into the aug matrix exactly

vividf:
## Summary

`ResizeCropFlipRotImage` reports an augmentation matrix that downstream
consumers use to update `camera_intrinsics` and project into the augmented
image, so the matrix must describe the pixel mapping exactly. It did not.

## The bug

The base scale used `min(final_height / source_height, final_width / source_width)`
on both axes. Whenever the source aspect ratio differed from `final_dim`, the
crop could not reach `final_dim` on one axis, so a second `cv2.resize` stretched
the pixels back up. That stretch entered the matrix only as the approximate
ratio `final_width / crop_width`, leaving the reported intrinsics inconsistent
with the image the model actually sees.

Using `max(...)` makes the base scale cover the crop on both axes, so one
resize plus one crop describes the mapping exactly and the matrix carries just
the resize scale and the crop offset.

## Behavior changes

Aligning the crop with the reference implementation changes two more things:

- The horizontal crop offset is random during training instead of always
  centered.
- A crop window reaching past the resized image is zero-padded instead of
  being stretched to fit.

## Impact on BEVFusion

This transform is reached through `ImageAug3D` in
`transforms/camera_lidar/camera_lidar.py`, so BEVFusion's image pipeline is
affected. Its nuScenes training config uses `resize_lim: [0.38, 0.55]`, whose
lower end previously triggered the hidden second resize. Please review with
that in mind: the augmented pixels and the reported intrinsics agree after
this change, but they are not identical to what BEVFusion trained on before.

## Testing

`tests/transforms/test_camera.py` adds a regression test that projects a known
pixel through the returned matrix and checks it lands on the same feature in
the augmented image, using a source aspect ratio that differs from
`final_dim` so the old code fails it.

## Context

Split out of #116 at review request, so the datamodule PR stays a single
concern.

## Stack
1. feat(datamodule): add scene-streaming dataloaders for temporal detectors — #116
2. **fix(transforms): fold ResizeCropFlipRotImage crop into the aug matrix exactly ← this PR**
3. feat(models): add StreamPETR building blocks — #118
4. feat(models): add the StreamPETR detector — [vividf#9](https://github.com/vividf/autoware-ml/pull/9)
5. feat(streampetr): add base nuScenes and T4 j6gen2 configurations — [vividf#10](https://github.com/vividf/autoware-ml/pull/10)
6. feat(streampetr): harden the streaming training infrastructure — [vividf#3](https://github.com/vividf/autoware-ml/pull/3)
7. feat(streampetr): project 2D annotations and fold augmentations into ego poses — [vividf#4](https://github.com/vividf/autoware-ml/pull/4)
8. feat(streampetr): add auxiliary 2D head, partial-ignore, and DDP loss normalization — [vividf#5](https://github.com/vividf/autoware-ml/pull/5)
9. feat(streampetr): pin the production three-stage training configs — [vividf#6](https://github.com/vividf/autoware-ml/pull/6)

#116, #117 and #118 are independent of each other and all based on `main`, so they can be reviewed and merged in any order. Each later layer is opened here once its prerequisites merge, so every PR stays a single-layer diff.


autoware_ml/transforms/camera/resize.py
```diff
@@ -223,44 +242,40 @@ def _augment_image(self, image: npt.NDArray) -> tuple[npt.NDArray[np.float32], n
...
+        crop_x = int(np.random.uniform(0, max_crop_x)) if self.training else max_crop_x // 2
+        cropped = _crop_with_zero_padding(resized, crop_x, crop_y, final_width, final_height)
 
         transform = np.eye(4, dtype=np.float32)
-        transform[0, 0] = resize * final_width / crop_width
-        transform[1, 1] = resize * final_height / crop_height
-        transform[0, 2] = -crop_x * final_width / crop_width
-        transform[1, 2] = -crop_y * final_height / crop_height
+        transform[0, 0] = resize
+        transform[1, 1] = resize
```
amadeuszsz: This is not valid anymore due to rounding via `int()`. You can recalculate the scale caused by rounding. Also consider whether you should first apply proper rounding and then cast to int instead of trimming the float value.

amadeuszsz:
@codex review

autoware_ml/transforms/camera/resize.py
vividf: Good catch, fixed in cc08e048.

amadeuszsz (approved):
LGTM

PR #119, opened 2026-09-08
feat(ptv3): export fused bf16 TransHead attention in all detection configs

mojomex:
## Summary

Enables the TensorRT-fusable TransHead attention export from #95 in every PTv3-family config that has a detection head, including the LitePT ones.

- set `model.bbox_head.use_bf16_cross_attention: true` and `deploy.onnx.modules.ptv3_det3d_head.precision: fp16`.
- The fp16 precision is set on the `ptv3_det3d_head` module only. The encoder and segmentation head exports keep their current precision, so this PR changes nothing but the detection head.
- Docs: `docs/models/ptv3.md` now states that the shipped configs export the fused path and shows the override to fall back to the non-fused fp32 head.
- Test: `tests/configs/test_deploy_metainfo.py` gains a composition test asserting that every detection config pairs the head flag with the fp16 module precision, and its config list now also covers the nuScenes detection config and both LitePT configs.

## Related Issues

- Follow-up to #95 (fused mixed-precision TransHead attention) and #89 (`deploy.onnx.precision`).
- Covers the LitePT configs added in #94.

## Checklist

### DIFF
--- a/autoware_ml/configs/tasks/detection3d/ptv3/voxel005_51m_nuscenes.yaml
+++ b/autoware_ml/configs/tasks/detection3d/ptv3/voxel005_51m_nuscenes.yaml
@@ -163,6 +163,8 @@ model:
     num_decoder_layers: 1
     num_heads: 8
     feedforward_channels: 256
+    # Fusable bf16 cross-attention export; requires the fp16 module precision below and SM80+.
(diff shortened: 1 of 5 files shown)


amadeuszsz (approved):
LGTM

PR #121, opened 2026-09-09
perf(lovasz): iterate the present classes as a host list

mojomex:
## Summary

`LovaszLoss._lovasz_softmax_flat` iterated `labels.unique()` directly. Iterating a CUDA tensor element by element synchronises with the device once per iteration (i.e. class).
Indexing `probabilities[:, class_index]` with the resulting 0-d CUDA tensor synchronises again, and the `foreground.sum() == 0` guard a third time. The guard is also redundant: every class that `unique` returns is present in `labels`.

The loop now iterates `labels.unique().tolist()`, a single device-to-host copy, and the guard is gone. Numerics are unchanged: the same operations run in the same order.

## Related Issues

Part of removing the per-step host synchronisations from PTv3 training; the large remainder (the attention padding loop) is addressed in a separate, independent PR.

## Checklist

## Additional Notes

The loss had no unit tests. `autoware_ml/tests/losses/test_lovasz.py` derives its expected values from the definition of Lovasz loss.

Also performed a gradient check. It exposed a latent dtype issue: `_lovasz_grad` hard-cast its cumulative sums to float32, so the loss could not run on float64 inputs at all. It now keeps the caller's dtype.

## Additional Log and Media

Counted with `torch.cuda.set_sync_debug_mode("warn")` on one training step of the tiny PTv3 lidarseg configuration (`segmentation3d/ptv3/voxel012_122m_t4dataset_j6gen2`): this loop accounted for 24 of the ~3060 host synchronisations per step (two per present class). After the change it accounts for one.

### DIFF
--- a/autoware_ml/losses/segmentation3d/lovasz.py
+++ b/autoware_ml/losses/segmentation3d/lovasz.py
@@ -134,10 +134,11 @@ class LovaszLoss(_Loss):
             return (probabilities * 0.0).sum()
 
         losses = []
-        for class_index in labels.unique():
+        # One device->host copy for the class list; iterating the CUDA tensor
+        # directly would synchronise once per class, and every class listed here
+        # is present, so the old per-class emptiness check was redundant.
+        for class_index in labels.unique().tolist():
             foreground = (labels == class_index).type_as(probabilities)
-            if foreground.sum() == 0:
-                continue
             class_errors = (foreground - probabilities[:, class_index]).abs()
             class_errors, permutation = torch.sort(class_errors, descending=True)
             foreground = foreground[permutation]
--- a/autoware_ml/losses/segmentation3d/lovasz.py
+++ b/autoware_ml/losses/segmentation3d/lovasz.py
@@ -36,8 +36,10 @@ def _lovasz_grad(sorted_ground_truth: torch.Tensor) -> torch.Tensor:
     """
     point_count = len(sorted_ground_truth)
     foreground_total = sorted_ground_truth.sum()
-    intersection = foreground_total - sorted_ground_truth.float().cumsum(0)
-    union = foreground_total + (1 - sorted_ground_truth).float().cumsum(0)
+    # Keep the caller's floating dtype rather than forcing float32, so the loss
+    # also runs in float64 (gradcheck) and mixed precision without a dtype clash.
+    intersection = foreground_total - sorted_ground_truth.cumsum(0)
+    union = foreground_total + (1 - sorted_ground_truth).cumsum(0)
     jaccard = 1.0 - intersection / union
     if point_count > 1:
         jaccard[1:point_count] = jaccard[1:point_count] - jaccard[0:-1]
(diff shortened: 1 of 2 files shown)


autoware_ml/losses/segmentation3d/lovasz.py
```diff
@@ -134,10 +134,11 @@ def _lovasz_softmax_flat(
             return (probabilities * 0.0).sum()
 
         losses = []
-        for class_index in labels.unique():
+        # One device->host copy for the class list; iterating the CUDA tensor
+        # directly would synchronise once per class, and every class listed here
+        # is present, so the old per-class emptiness check was redundant.
```
mojomex:
```suggestion
        # `.tolist()` synchronized once. Omitting it would sync per iteration.
```

autoware_ml/tests/losses/test_lovasz.py
```diff
@@ -0,0 +1,98 @@
...
+    losses = []
+    for class_index in labels.unique():
+        foreground = (labels == class_index).type_as(probabilities)
+        if foreground.sum() == 0:
+            continue
+        class_errors = (foreground - probabilities[:, class_index]).abs()
+        class_errors, permutation = torch.sort(class_errors, descending=True)
+        foreground = foreground[permutation]
+        losses.append(torch.dot(class_errors, _lovasz_grad(foreground)))
+    return torch.stack(losses).mean()
```
mojomex: The new implementation is trivially valid if the old one was valid. Don't treat the old impl as ground truth - instead focus on mathematically correct behavior regardless of impl.

autoware_ml/tests/losses/test_lovasz.py
```diff
@@ -0,0 +1,98 @@
...
+    (grad,) = torch.autograd.grad(loss, logits)
+
+    ref_logits = logits.detach().requires_grad_(True)
+    valid = labels != IGNORE
+    ref_loss = _reference_lovasz_softmax_flat(ref_logits.softmax(dim=1)[valid], labels[valid])
+    (ref_grad,) = torch.autograd.grad(ref_loss, ref_logits)
+
+    assert torch.allclose(loss, ref_loss, rtol=1e-6, atol=0)
+    assert torch.allclose(grad, ref_grad, rtol=1e-5, atol=1e-8)
+    assert torch.all(grad[~valid] == 0)
```
mojomex: Same comment as above.

autoware_ml/tests/losses/test_lovasz.py
mojomex: Ideally there would be more hand-crafted tests to cover all edge and normal cases.

autoware_ml/losses/segmentation3d/lovasz.py
mojomex: Applied verbatim.

autoware_ml/tests/losses/test_lovasz.py
mojomex: Agreed, and done in 08917eeb354b179c02947e85de357575f9dab8a9: the reference loop is gone. The tests now derive expected values from the definition: a hand-computed two-class example worked through in the docstring, perfect prediction = 0, completely wrong prediction = 1, the [0, 1] bound, absent classes excluded from the mean, ignored points affecting neither the value nor receiving gradient, invariance to point order, a finite-difference gradcheck in float64, the all-ignored batch, and per_image / loss_weight. The gradcheck exposed a latent dtype issue: `_lovasz_grad` hard-cast to float32, so the loss could not run in float64 at all. It now keeps the caller's dtype.

mojomex: Added, see the reply above for the list.

PR #122, opened 2026-09-09
perf(ptv3): build the attention padding for the whole batch at once

mojomex:
## Summary

`SerializedAttention._get_padding_and_inverse` looped over the samples of a batch in Python and sliced CUDA tensors with 0-d CUDA bounds (`offset[i]`, `padded_offset[i + 1]`, ...). Every such bound is a device-to-host copy, so each attention block paid about 15 synchronisations per sample, and every block of a stage recomputed the same result. On a batch of 16 that is roughly 2900 synchronisations per training step, each of which drains the GPU queue, so the CPU could never run ahead of the GPU and the GPU sat idle between them.

This PR

- builds the padded token grid for the whole batch in one pass (`searchsorted` for the sample of every slot and token, then the same "borrow the previous window" rule as before, expressed as a `where`), which costs two synchronisations per call instead of ~240;
- caches `pad`, `unpad` and `cu_seqlens` on the `Point` per patch size, as the reference PTv3 implementation does, so a stage computes them once rather than once per block (a new `Point` is created at every pooling, which invalidates the cache; export mode never caches).

The result is bit-identical to the loop it replaces.

## Related Issues

Part of removing the per-step host synchronisations from PTv3 training; a small companion change to the Lovasz loss is in https://github.com/tier4/autoware-ml/pull/121.

## Checklist

## Additional Notes

`test_batched_padding_matches_the_per_sample_reference` keeps the original loop as the reference and checks `pad`, `unpad` and `cu_seqlens` for exact equality over 8 random batches x 3 patch sizes, each batch containing a sample that is an exact multiple of the window, one below one window, and an empty sample. The existing export-fill tests keep passing, so export behaviour is unchanged.

## Additional Log and Media

Measured on the tiny PTv3 lidarseg configuration (`segmentation3d/ptv3/voxel012_122m_t4dataset_j6gen2`, AMP off, batch 16, 209 steps, RTX PRO 6000, dataset on a 10 Gbit NAS so the loader is not the bottleneck), same seed, together with https://github.com/tier4/autoware-ml/pull/121:

| | before | after |
| --- | ---: | ---: |
| host synchronisations per training step | 3062 | 89 |
| steady-state step time | 1270 ms | 1147 ms |
| encoder forward + backward per step | 929 ms | 758 ms |
| one epoch, wall time | 401 s | 348 s |
| GPU busy per step (torch.profiler) | 944 ms | 1002 ms |
| val/loss after one epoch | 1.2557 | 1.2537 |

GPU work per step is unchanged within noise; the saving is CPU stall time. Synchronisations counted with `torch.cuda.set_sync_debug_mode("warn")`, attributed to the innermost `autoware_ml` frame.


autoware_ml/models/segmentation3d/encoders/ptv3.py
```diff
@@ -577,40 +577,26 @@ def _get_padding_and_inverse(
...
-                        step=self.patch_size,
-                        dtype=torch.int32,
-                        device=point.offset.device,
-                    )
-                )
-        if cu_seqlens is None:
+        # Build the padded token grid for the whole batch at once. The per-sample
+        # loop this replaces sliced CUDA tensors with 0-d CUDA bounds, and every
+        # bound cost a host synchronisation: ~15 per sample per attention block,
+        # about 3000 per training step, which serialised the CPU against the GPU.
```
mojomex:
```suggestion
        # Build the padded token grid for the whole batch at once. This minimizes
        # the amount of host synchronizations needed.
```

autoware_ml/models/segmentation3d/encoders/ptv3.py
```diff
@@ -577,40 +577,26 @@ def _get_padding_and_inverse(
...
+        device = point.offset.device
```
mojomex, line 582: What about DDP / multiple devices?

autoware_ml/models/segmentation3d/encoders/ptv3.py
mojomex: Applied in 7070f4b.

autoware_ml/models/segmentation3d/encoders/ptv3.py
mojomex: Both are unaffected, and the device handling is the same as in the loop this replaces (it also created its tensors on `point.offset.device`).

- **Device**: nothing here uses the current/default device. `device` is read off the incoming batch, and every tensor in the function either passes it explicitly or is derived from an input tensor, so under DDP each rank builds its grid on the device its own shard already lives on.
- **DDP**: the new cache lives on the `Point`, which is created inside the forward and thrown away with the batch, so it is per-process and per-step - there is no state shared between ranks and nothing that survives into the next step. It also never skips a module call: `qkv` and `proj` still run in every block, only the recomputation of the index tensors is skipped, so the reducer still sees every parameter.

Checked rather than assumed:

- `pad`, `unpad` and `cu_seqlens` for the same batch on CPU and on CUDA are equal element-wise and each lands on the device of its input (patch sizes 4/16/48, flash on and off).
- A 2-rank gloo DDP run with different batch shapes per rank - so the ranks take different window sizes and different cache keys (`pad_24` on rank 0, `pad_8` on rank 1) - completes the backward pass and ends with identical all-reduced gradients on both ranks.

One caveat: this machine has a single GPU, so the two-GPU case is covered by the device-parity check above and by reading the code, not by an actual 2-GPU nccl run.

PR #124, opened 2026-09-11
chore: update pixi to 0.80.0 and t4-devkit to 0.8.0

amadeuszsz:
## Summary

The pixi install layer of the Docker build times out on some machines with 0.66.0. Newer pixi releases look to have addressed that, which is the main reason for the bump.

Bumps pixi 0.66.0 to 0.80.0 across the docs workflow, ansible role, Dockerfile and the installation guide. 0.80 replaces `[tool.pixi.system-requirements]` with per-platform entries, so `platforms` now carries the cuda and glibc requirements directly. The lock moves to format v7, which renames the platform key and accounts for the whole diff. No package version changed in that migration.

Bumps t4-devkit 0.6.0 to 0.8.0 in the same pass. The devkit renamed `Tier4` to `T4Devkit`, so the record generator and the class diagram follow the new name. The rest of the devkit surface the generator uses is unchanged. The lock picks up `pypcd4` and `python-neo-lzf`, which the devkit now needs for PCD point clouds.

## Checklist

## Additional Notes

The image has to be rebuilt, since the pixi version is baked in at build time.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -26,7 +26,7 @@ jobs:
       - uses: actions/checkout@v6
       - name: Install pixi
         env:
-          PIXI_VERSION: 0.66.0
+          PIXI_VERSION: 0.80.0
         run: |
           export PIXI_HOME="$HOME/.pixi"
           export PIXI_BIN_DIR="$HOME/.pixi/bin"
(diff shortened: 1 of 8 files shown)


KSeangTan (approved):
LGTM!

PR #127, opened 2026-09-15
refactor(tests)!: colocate the test suites with the packages they cover

amadeuszsz:
## Summary

Test suites move next to the packages they cover, so a package carries its own `tests/` directory instead of the whole tree living under one top-level `autoware_ml/tests`.

- 90 file moves, no test logic touched. `autoware_ml/tests/metrics` becomes `autoware_ml/metrics/tests`, `autoware_ml/tests/models` splits into `models/detection3d/tests`, `models/multi/tests` and `models/segmentation3d/tests`, and the rest follow the same rule.
- The only content edits are 10 import lines pointing at the moved fixtures and conftest modules, plus one path in the LitePT doc table.
- `autoware_ml/tests` is gone. Nothing changes for callers, `pytest autoware_ml/` already collects the new locations, so neither the pixi task nor the CI workflow needs an update.

## Checklist

- [x] You've checked our [contribution overview](https://tier4.github.io/autoware-ml/contributing/contribution-overview/).
- [x] Your PR follows our [pull request guidelines](https://tier4.github.io/autoware-ml/contributing/pull-request-guidelines/).

## Additional Notes

Review is easiest with rename detection on, `git show -M` renders it as 90 renames and 11 small content edits.

## Additional Log and Media

`pytest autoware_ml`: 898 passed at this branch head, same count as on #112.

PR #129, opened 2026-09-15
refactor(utils): type the array signatures of the calibration, dataset and spconv helpers

amadeuszsz:
## Summary

Array parameters in the shared helpers say their shape and dtype instead of only their dtype, using the `jaxtyping` annotations already used across the datamodule.

- `CalibrationData` fields become `Float32[np.ndarray, "3 3"]`, `Float32[np.ndarray, "4 4"]` and `Float32[np.ndarray, " num_coefficients"]`, so the shapes the docstring described in prose are now in the signature and the prose goes.
- `convert_quaternion_to_matrix` states the translation is a 3-vector and the result a 4x4, in both the annotation and the docstring.
- The two spconv `ImplicitGemm` entry points type their `masks` list as `UInt32[np.ndarray, " 1"] | None` rather than a bare `NDArray`.

## Related Issues

Stacked on #127.

## Checklist

## Additional Notes

- Annotations only, no runtime behaviour changes. `jaxtyping` is already a dependency and already used elsewhere in the tree.
- Leading whitespace is forced by type checker, jaxtyping effectively will ignore this whitespace.

## Additional Log and Media

`pytest autoware_ml`: 898 passed at this branch head, same count as on #127.


KSeangTan (dismissed):
LGTM! However, i would suggest to replace `list` by `Sequence` for immutability as well.

PR #133, opened 2026-09-16
feat(taxonomy)!: database owned label vocabulary and taxonomy levels with class keyed tables

amadeuszsz:
## Summary

The label vocabulary and the trained class set move into the database, so the corpus, the models and the metrics read one table instead of each config repeating its own class list.

- `LabelVocabulary` names every raw category the corpora carry and folds it onto a fine label. A raw category outside every level is listed with `null`, a raw category missing from the list fails the table generation.
- `LabelTaxonomy` and its `DetectionTaxonomy` / `SegmentationTaxonomy` subclasses coarsen those fine labels onto the classes a level trains, and carry the class keyed tables next to them: eval range, collision kind, living run speed, partial detection classes, heatmap pooling classes and class groups. Instantiation validates that every fine label folds onto one class or `null` and that each table names exactly the classes of the level.
- T4dataset gets two levels. `online` is the 10 detection / 16 segmentation set the deployed models train, `offline` the finer set the pseudo labelling produces. The segmentation class list starts with the detection classes in the same order, so a class shares one index across both tasks.
- nuScenes gets the same structure over its existing 7 detection / 15 segmentation classes.
- `CollisionTTC` takes `kinds` and `living_speeds` from that taxonomy instead of its own module level defaults, so the collision model and the trained classes cannot drift apart.

## Related Issues

Stacked on #129.

## Checklist

## Additional Notes

- This PR breaks behaviour on its own and a follow-up PR in this stack fixes it. `CollisionTTC` now requires `kinds` and `living_speeds`, but the T4 dataset config still builds it without them, so the T4 detection configs do not instantiate at this commit. The same follow-up points `class_names` at the taxonomy, which is what makes the shortened `score_threshold` lists here line up with the head.
- This PR touches BEVFusion, CenterPoint and TransFusion configs - every model currently shares one `online` taxonomy and the class sets are aligned across them. Later on the `online` level will be split into separate taxonomies. Currently developed model like PTv3, BEVFusion and StreamPETR can got their own taxonomies, rest of the models can share common taxonomy.

## Additional Log and Media

`pytest`: 889 passed at this branch head.


autoware_ml/configs/database/t4dataset/taxonomy/vocabulary/default_detection3d.yaml
```diff
@@ -0,0 +1,98 @@
+# Detection vocabulary of T4dataset: every raw box category the corpora carry, mapped onto the
+# fine label names the levels coarsen. A raw category outside every level is listed with null,
+# a raw category missing from the list fails the table generation. Box records store the fine
+# name whatever level they were baked for.
+_target_: autoware_ml.databases.taxonomy.LabelVocabulary
+_convert_: all
+
+name_mapping:
```
KSeangTan: Are we gonna replace 
`autoware_ml/configs/database/t4dataset/database_task_configs/detection3d/labels/default_label_remapper.yaml `

by vocabulary?

KSeangTan: If that's the case, i would like add `default_` as a prefix name.

autoware_ml/configs/database/t4dataset/taxonomy/offline.yaml
```diff
@@ -0,0 +1,230 @@
+# Offline taxonomy of T4dataset, the L1 layer of the pseudo labeling taxonomy: every online class
```
KSeangTan: I dont think offline/online are a good name, i would consider to use `pseudo_label` or `manual_label`.

autoware_ml/configs/database/t4dataset/taxonomy/offline.yaml
```diff
@@ -0,0 +1,230 @@
...
+# splits into named subtypes trained by the teacher models used for pseudo labeling and
+# distillation, never deployed. A trailer keeps a box of its own, so the databases bind it
+# together with the pipelines that do not merge trailers.
+_target_: autoware_ml.databases.taxonomy.DatabaseTaxonomy
+_convert_: all
+
+defaults:
```
KSeangTan: This file is too long, please split them to different tasks.

autoware_ml/configs/database/t4dataset/taxonomy/offline.yaml
```diff
@@ -0,0 +1,230 @@
...
+    traffic_cone: static
+    debris.unclassified: static
+    debris.pushable_pullable: static
+  living_speeds:
+    bicycle: 6.0
+    pedestrian.common: 3.0
+    pedestrian.personal_mobility: 6.0
+    pedestrian.stroller: 3.0
+    animal: 4.0
+  partial_detection_classes:
```
KSeangTan: Anything related to model shouldn't be part of this file.

autoware_ml/configs/database/t4dataset/taxonomy/offline.yaml
```diff
@@ -0,0 +1,230 @@
...
+    [motorcycle, bicycle, pedestrian.common, pedestrian.personal_mobility, pedestrian.stroller,
+     traffic_cone]
+  heatmap_pooling_classes:
```
KSeangTan: Anything related to model shouldn't be part of this file.

autoware_ml/configs/database/t4dataset/taxonomy/offline.yaml
```diff
@@ -0,0 +1,230 @@
...
+    [car.common, car.emergency, car.special, truck.common, truck.emergency, truck.trailer,
+     truck.construction, bus, train]
+  class_names:
+    - car.common
```
KSeangTan: I dont see the reason why we need to define these?

autoware_ml/configs/database/t4dataset/taxonomy/offline.yaml
```diff
@@ -0,0 +1,230 @@
...
+    - pedestrian.common
+    - pedestrian.personal_mobility
+    - pedestrian.stroller
+    - animal
+    - barrier.movable
+    - barrier.construction_sign
+    - traffic_cone
+    - debris.unclassified
+    - debris.pushable_pullable
+  coarsening:
```
KSeangTan: I am confused between this and name_mapping.

autoware_ml/metrics/tests/test_collision_adapter.py
```diff
@@ -12,13 +12,32 @@
 from autoware_ml.metrics.tests.conftest import CLASS_NAMES, EGO, FakeMapProvider, collision_box
 from autoware_ml.types.metrics import AgentKind
 
+KINDS = {
```
KSeangTan, line 15: `behavior` will be a better name than `kinds`.

KSeangTan (changes requested):
LGTM overall, however, i have few questions about the mapping.

KSeangTan:
@codex review

autoware_ml/configs/database/t4dataset/taxonomy/vocabulary/default_detection3d.yaml
amadeuszsz: Yes. The vocabulary replaces default_label_remapper.yaml and the class list replaces default_class_names.yaml. Both files are deleted in the follow-up PR that binds the databases to the taxonomy, this one only adds the new layer.

amadeuszsz: Agreed, they are default_detection3d.yaml and default_segmentation3d.yaml now. Addressed in https://github.com/tier4/autoware-ml/commit/0be61e46

autoware_ml/configs/database/t4dataset/taxonomy/offline.yaml
amadeuszsz: They do not say how the labels were produced, both levels are used with manual and with pseudo labels, so pseudo_label and manual_label would name the wrong axis. What differs is granularity, so they are default_coarse and default_fine now. The fine set is what the large models train, the coarse set is what the production model runs. A variant of either drops the prefix and names itself. Addressed in https://github.com/tier4/autoware-ml/commit/0be61e46

autoware_ml/configs/database/t4dataset/taxonomy/offline.yaml
amadeuszsz: Split per task. The levels live in taxonomy/detection3d and taxonomy/segmentation3d next to the vocabularies, and the top level file only composes the two. The fine level went from 227 lines to 90 and 140. Addressed in https://github.com/tier4/autoware-ml/commit/0be61e46

autoware_ml/configs/database/t4dataset/taxonomy/offline.yaml
amadeuszsz: Agreed, it is a metric input rather than a taxonomy one, so it moved out to the metric that reads it. Addressed in https://github.com/tier4/autoware-ml/commit/40db109e

autoware_ml/configs/database/t4dataset/taxonomy/offline.yaml
amadeuszsz: Agreed, moved out to the model configs. It does depend on the taxonomy, since by default it covers the larger objects, but the model is the right place to decide it. Addressed in https://github.com/tier4/autoware-ml/commit/40db109e

autoware_ml/configs/database/t4dataset/taxonomy/offline.yaml
amadeuszsz: class_names is the trained class set of the level, and its order is the label index order: index 0 is car.common and so on. The boxes are baked with those indices and the model heads size themselves from this list, so it cannot be derived from the coarsening, which is unordered.

autoware_ml/configs/database/t4dataset/taxonomy/offline.yaml
amadeuszsz: They work at different levels. name_mapping is dataset level: it renames every raw category the corpus carries into one standard name, applies to all data in that dataset, and lives in one place, so a corpus fix is made once. coarsening works on those standard names and defines the taxonomy a model trains. We will have several taxonomies, BEVFusion will want its own, and they all coarsen the same standard names, which keeps each of them short and readable and keeps the dataset level renaming out of them.

autoware_ml/metrics/tests/test_collision_adapter.py
amadeuszsz: The name comes from main, CollisionTTC already takes kinds and the enum is AgentKind, 54 uses across the metrics. This PR only makes the argument required and adds the table that feeds it. I checked every Kind name in the tree and it is consistent, so I would keep it for now and rename the whole thing at the end of this PR chain rather than half of it here.

autoware_ml/metrics/detection3d/collision.py
```diff
@@ -105,20 +87,19 @@ def __init__(
         map_provider: LaneletMapProvider,
         *,
         vehicle: VehicleGeometry,
+        kinds: Mapping[str, AgentKind | str],
+        living_speeds: Mapping[str, float],
```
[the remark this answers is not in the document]
amadeuszsz: Correct and intended. The dataset config is wired to the taxonomy in the follow-up PR in this stack, which is also what makes the shortened threshold lists line up. It is noted in the PR description.

autoware_ml/configs/tasks/detection3d/bevfusion/lidar_voxel0170_second_secfpn_120m_t4dataset_j6gen2.yaml
```diff
@@ -148,7 +148,7 @@ model:
         post_max_size: 50
     bbox_coder:
       post_center_range: [-200.0, -200.0, -10.0, 200.0, 200.0, 10.0]
-      score_threshold: [0.015, 0.010, 0.010, 0.010, 0.020, 0.020, 0.030, 0.030, 0.020, 0.040, 0.020, 0.020, 0.020]
+      score_threshold: [0.015, 0.010, 0.010, 0.020, 0.020, 0.030, 0.030, 0.020, 0.040, 0.020]
```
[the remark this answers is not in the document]
amadeuszsz: Same follow-up. The class list still comes from the old 13 class dataset config at this commit, so the thresholds only line up once that config reads the taxonomy. The intermediate commits of this stack are not meant to run.

autoware_ml/configs/tasks/detection3d/bevfusion/lidar_voxel0170_second_secfpn_120m_t4dataset_j6gen2.yaml
```diff
@@ -116,7 +116,7 @@ model:
     nms_type: circle
     nms_min_radius: 0.5
     nms_groups:
-      - class_names: [car, truck, bus, train]
+      - class_names: [car, truck, bus]
```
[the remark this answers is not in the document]
amadeuszsz: Same cause as the threshold list. train leaves the taxonomy and the NMS group in one step, and the dataset config that still declares it is switched in the follow-up PR.

autoware_ml/databases/taxonomy.py
```diff
@@ -0,0 +1,466 @@
...
+
+    @property
+    def fine_names(self) -> tuple[str, ...]:
+        """Fine label names of the vocabulary, sorted."""
+        return self._fine_names
+
+    @property
+    def name_mapping(self) -> Mapping[str, str | None]:
+        """Raw label name to fine label name, None for a raw label outside every level."""
+        return self._name_mapping
```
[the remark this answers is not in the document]
amadeuszsz: Good catch, that one is real. Every dictionary accessor hands out a MappingProxyType now, so the hash a table was baked with cannot change under a consumer. There is a test for it. Addressed in https://github.com/tier4/autoware-ml/commit/741daa0c

autoware_ml/configs/database/t4dataset/taxonomy/offline.yaml
KSeangTan: Okay, i got it, basically:

`annotation_name -> name_mapping`, any label not listed in `name_mapping` is kept as their original annotation_name. 

coarsening is related to `class_names`, where each annotation_name/name_mapping is mapped to `class_names` based on this mapping.
 
In this case, `coarsening` can be a bit confusing, I would recommend to change it to `class_name_mapping`, which is more straightforward.

KSeangTan (dismissed):
LGTM overall, please address the comment and feel free to merge it.

autoware_ml/configs/database/t4dataset/taxonomy/offline.yaml
amadeuszsz: Ok, I looked on it again and I think I have a better convention:
- `name_mapping` becomes `class_renaming`. It operates on dataset raw labels, has no model knowledge, it is not a mapping, just an unified naming convention. Shared my all the taxonomies/models.
- `coarsening` becomes `class_mapping`. This is what your model will consider. It is mapping because here you potentially map objects semantically different.
- `class_names` enumerates the classes.

So first `class_renaming` standardizes vocabulary, then `class_mapping` does the mapping, then `class_names` defines the order.

amadeuszsz: Addressed in https://github.com/tier4/autoware-ml/pull/133/commits/57ca790108c222d0de01070221052ca8ada95114 (about applied convention) and https://github.com/tier4/autoware-ml/pull/133/commits/7dc364772e274758708f91923fd81a045bfaef27 (about refinement of semseg related names due to naming collision).

PR #137, opened 2026-09-17
refactor(cli)!: drop the dataset info file entrypoint and share the workspace paths

amadeuszsz:
## Summary

The dataset info file path goes, and the workspace directories every entrypoint needs are defined once.

- `create-dataset`, `scripts/create_dataset.py` and the whole `autoware_ml/tools/dataset/` tree are deleted. The pkl info files they produced are not read anywhere: the datamodule reads the parquet record table a database generates, and training generates a missing table itself.
- `configs/defaults/modules/paths.yaml` holds `data_root_path`, `cache_root_path` and `working_dir`, and the generator configs include it instead of restating them.
- The data directory is mounted read only, in `docker/container.sh` and in the devcontainer. Nothing the framework writes belongs there.
- `cache_root_path` moves to `${working_dir}/.cache` as a consequence, so the record tables land on the read-write mount.
- `data_root_path` reads `AUTOWARE_ML_DATA_PATH` rather than the literal `data`, which is the variable the container already mounts from.

## Related Issues

First of the chain that replaces the multi task dataset path with the typed one.

## Checklist

## Additional Notes

- Almost all deletion, 59 lines added against 1298 removed.
- The only behaviour change is the read only data mount and the cache directory that follows from it. Regenerating a record table now writes to `${working_dir}/.cache` instead of below the data directory, so an existing table is not found and is generated again in the new place.
- `scripts/generate_dataset.py` stays and keeps its name. It loads from the bundled config root through `get_config_path()`, the same way `train.py` already does, which is why the generator configs gain `# @package _global_` and reference their groups absolutely.

## Additional Log and Media

`pytest`: 887 passed at this branch head. Both generator configs compose, with `data_root_path=/workspace/data` and `cache_root_path=/workspace/.cache`.


autoware_ml/configs/defaults/modules/paths.yaml
```diff
@@ -0,0 +1,9 @@
+# @package _global_
+
+# Root of the data mount the record paths resolve against.
+data_root_path: ${oc.env:AUTOWARE_ML_DATA_PATH}
+# Directory the record tables are written to. They index the datasets and stay small, so
+# they live below the workspace, which is mounted read-write while the data mount is read only.
+cache_root_path: ${working_dir}/.cache
+# Directory holding the perception-devops checkout with the scenario lists.
```
KSeangTan: Remove `perception-devops`

KSeangTan (dismissed):
Thanks! LGTM!

autoware_ml/configs/defaults/modules/paths.yaml
amadeuszsz: Fixed 👍🏻

PR #136, opened 2026-09-17
feat(dataclasses): geometry samples and detection3d GT batch

KSeangTan:
## Summary

PR 2 of the stacked chain that splits `feat/add_multitask_bevfusion_experiments` into reviewable pieces. Stacked on #135 (PR 1), which provides `BaseImages`;

This PR introduces the `autoware_ml.dataclasses` package, the typed replacement for `datamodule/multi_task/dataclasses`:

- `dataclasses/geometry/images.py`: `ImageSample` and `ImageGTBatch`, collating a sequence of `BaseImages` into batched images, intrinsics, `lidar2images`, `lidar2cams` and augmentation matrices.
- `dataclasses/geometry/point_clouds.py`: `LiDARPointCloudSample` and `PointCloudGTBatch`, collating `BasePoints` with batch indices.
- `dataclasses/batch/detection3d.py`: `Detection3DGTSample` and `Detection3DGTBatch`, the padded ground-truth containers (`gt_bboxes_3d`, `gt_labels_3d`, `gt_valid_bboxes`, `gt_bboxes_num_points`) every detection head consumes downstream.
- `dataclasses/geometry/transformation.py` and `dataclasses/batch/segmentation3d.py`: byte-identical copies of the current `datamodule/multi_task/dataclasses` modules, so the package is complete. The originals stay until the datamodule PR removes `datamodule/multi_task`.

Purely additive: nothing on `main` imports the new package yet. The next PR adds `ModelGTSample`/`ModelGTBatch` on top of these containers, together with the `dataclasses/batch/tests` suite that exercises their collation; the containers here carry no logic beyond field validation and concatenation.

## Related Issues

Part of the split of `feat/add_multitask_bevfusion_experiments`. Depends on: #135.

## Checklist

- [x] You've checked our [contribution overview](https://tier4.github.io/autoware-ml/contributing/contribution-overview/).
- [x] Your PR follows our [pull request guidelines](https://tier4.github.io/autoware-ml/contributing/pull-request-guidelines/).

## Additional Notes

Budget: 497 non-test lines added, 0 deleted. Files are byte-identical to the feature branch.

Verification: all new modules import alongside the stdlib `dataclasses` module; `pytest autoware_ml --collect-only` collects 898 tests with no errors; `pytest autoware_ml/geometry autoware_ml/datamodule/tests` passes (72 tests); ruff check/format clean.

## Additional Log and Media

```
898 tests collected in 4.71s
72 passed in 1.61s
```

PR #138, opened 2026-09-17
feat(dataclasses): add ModelGTSample, ModelGTBatch and frame meta

KSeangTan:
## Summary

PR 3 of the stacked chain that splits `feat/add_multitask_bevfusion_experiments` into reviewable pieces. Stacked on #136 (PR 2).

Adds the task-agnostic ground-truth containers that every typed transform, dataset and model in the chain will consume:

- `dataclasses/batch/sample_batch.py`: `ModelGTSample` (one frame: point cloud, images, detection3d and segmentation3d GT, frame meta, IO timing) and `ModelGTBatch`, with typed collation helpers (`collate_pointcloud_gt_samples`, `collate_detection3d_gt_samples`, `collate_image_gt_samples`, `collate_frame_meta_samples`, `collate_gt_samples`), `to_device()` and `infer_batch_size()`. This is the typed replacement for the string-keyed `CollationStrategy` map in `datamodule/collation.py`, which stays in place until the datamodule PR.
- `dataclasses/batch/frame_meta.py`: `FrameMetaSample` and `FrameMetaBatch` carrying `ego2global` and the scene token. Collation composes the inverse of the lidar augmentation so downstream map-based metric filters see the pose of the augmented frame.
- `dataclasses/batch/tests/test_sample_batch.py`: collation, device transfer and batch-size inference tests for the above.
- `pyproject.toml`: enables `consider_namespace_packages` for pytest. Tests colocated under `autoware_ml/dataclasses/**/tests` would otherwise be imported as `dataclasses.batch.tests...` and shadow the standard library `dataclasses` module.

Purely additive: nothing on `main` imports the new containers yet.

## Related Issues

Part of the split of `feat/add_multitask_bevfusion_experiments`. Depends on: #136. 

## Checklist

## Additional Notes

Budget: 398 non-test lines added (of which 7 are the `pyproject.toml` block), 256 test lines. Python files are byte-identical to the feature branch.

Verification: `pytest autoware_ml/dataclasses` passes (13 tests); `pytest autoware_ml --collect-only` collects 927 tests with no errors; ruff check/format clean.

## Additional Log and Media

```
13 passed in 0.59s
927 tests collected in 4.69s
```


autoware_ml/dataclasses/batch/sample_batch.py
```diff
@@ -0,0 +1,288 @@
...
+
+    def infer_batch_size(self) -> Int32:
+        """
+        Infer the batch size from the collated multi-task GT batch.
+
+        Returns:
+            Batch size if it can be inferred, otherwise raises ValueError.
+        """
+        if self.point_cloud_gt_batch is not None:
+            return torch.max(self.point_cloud_gt_batch.batch_indices) + 1
```
amadeuszsz: What if last sample in the batch has no points? I already had this issue before and applied similar method but this can lead to hard to debug issue. I think we should infer batch size during data collation and preserve it through device transfer. This change may affect models implementation, so you are free to leave TODO and fix it in follow up PR.

autoware_ml/dataclasses/batch/sample_batch.py
```diff
@@ -0,0 +1,288 @@
...
+
+        # Check if detection3d_gt_bboxes_3d are available in the samples
+        available_camera_image_data = gt_samples[0].camera_image_data is not None
+        if not available_camera_image_data:
+            return None
+
+        image_gt_samples = []
+        for sample in gt_samples:
+            if sample.camera_image_data is None:
+                raise ValueError("All samples must have camera_image_data for collating.")
```
amadeuszsz: 1. First you check if 1st sample in the batch has `camera_image_data` and if so, you return None.
2. Then you check all the samples in the batch and if so, you raise an error.

This gives you different behavior depends on shuffle order.

autoware_ml/dataclasses/batch/frame_meta.py
```diff
@@ -0,0 +1,103 @@
...
+
+        ego2globals = []
+        for frame_meta, lidar_transformation in zip(
+            frame_meta_samples, lidar_transformation_samples
+        ):
+            ego2global = frame_meta.ego2global
+            if lidar_transformation is not None:
+                ego2global = ego2global @ torch.linalg.inv(
+                    lidar_transformation.transformation_matrix.to(ego2global)
+                )
```
amadeuszsz, line 84: Why you undo transformation? Potentially it is harmful for metrics and calculating the collision area. However, in test set we do not augment so it should not occur.

amadeuszsz (changes requested):

autoware_ml/dataclasses/batch/sample_batch.py
```diff
@@ -0,0 +1,288 @@
...
+
+        Returns:
+          PointCloudGTBatch: Collated point cloud GT batch.
+        """
+        if len(gt_samples) == 0:
+            return None
+
+        available_pointcloud = gt_samples[0].point_cloud_data is not None
+        if not available_pointcloud:
+            return None
```
amadeuszsz: Similar comment as before: `[no points, points]` returns None and silently drops the second sample's cloud, while `[points, no points]` raises at line 145.

autoware_ml/dataclasses/batch/sample_batch.py
```diff
@@ -0,0 +1,288 @@
...
+        Returns:
+          Detection3DGTBatch: Collated detection3d GT batch.
+        """
+        if len(gt_samples) == 0:
+            return None
+
+        # Check if detection3d_gt_bboxes_3d are available in the samples
+        available_detection3d_gt_bboxes_3d = gt_samples[0].detection3d_gt_bboxes_3d is not None
+        if not available_detection3d_gt_bboxes_3d:
+            return None
```
amadeuszsz: Same issue - depending on 1st sample.

autoware_ml/dataclasses/batch/sample_batch.py
```diff
@@ -0,0 +1,288 @@
...
+        Returns:
+          ImageGTBatch: Collated images GT batch.
+        """
+        if len(gt_samples) == 0:
+            return None
+
+        # Check if detection3d_gt_bboxes_3d are available in the samples
+        available_camera_image_data = gt_samples[0].camera_image_data is not None
+        if not available_camera_image_data:
+            return None
```
amadeuszsz: Same issue - depending on 1st sample.

autoware_ml/dataclasses/batch/frame_meta.py
KSeangTan: I was thinking it's important to undo the transformation since we might use it in the training pipeline.

autoware_ml/dataclasses/batch/sample_batch.py
KSeangTan: Make sense, I have removed first sample checking, and it will raise an error if any sample don't have required attributes:
ccee996bc8c568ad5445a4b04dae3fd5c49d342c

autoware_ml/dataclasses/batch/sample_batch.py
KSeangTan: Thanks, updated here:
ccee996bc8c568ad5445a4b04dae3fd5c49d342c

autoware_ml/dataclasses/batch/sample_batch.py
KSeangTan: Thanks, updated here:
ccee996bc8c568ad5445a4b04dae3fd5c49d342c

autoware_ml/dataclasses/batch/sample_batch.py
KSeangTan: Thanks, updated here:
ccee996bc8c568ad5445a4b04dae3fd5c49d342c

autoware_ml/dataclasses/batch/sample_batch.py
KSeangTan: Fair point, I fixed it here:
613e9a51da0301243f34556a3a144382b98de10d

autoware_ml/dataclasses/batch/frame_meta.py
KSeangTan: Add unit tests here:
174d549ca6ef0f46594332ef3e64a167157d68b1

amadeuszsz (approved):
LGTM

